PR #3, opened 2024-06-06
Validation logic for Quark Builder

kevincheng96:


src/builder/BridgeRoutes.sol
```diff
@@ -0,0 +1,465 @@
...
+    string constant VERSION = "1.0.0";
+
+    string constant PAYMENT_METHOD_OFFCHAIN = "OFFCHAIN";
+    string constant PAYMENT_METHOD_PAYCALL = "PAY_CALL";
+    string constant PAYMENT_METHOD_QUOTECALL = "QUOTE_CALL";
+
+    string constant ACTION_TYPE_BRIDGE = "BRIDGE";
+    string constant ACTION_TYPE_TRANSFER = "TRANSFER";
+
+    string constant PAYMENT_CURRENCY_USD = "usd";
```
fluffywaffles, line 18: Probably no need for payment currencies that are not tokens.

src/builder/BridgeRoutes.sol
```diff
@@ -0,0 +1,465 @@
...
+        uint256 balance;
+    }
+
+    struct Payment {
+        // TODO: rename?
+        bool isToken;
+        string currency;
+        // TODO: make into struct?
+        uint256[] chainIds;
+        uint256[] maxCosts;
```
fluffywaffles, line 70: may as well, simply for the sake of consistency.

src/builder/BridgeRoutes.sol
```diff
@@ -0,0 +1,465 @@
...
+
+    struct BridgeActionContext {
+        uint256 amount;
+        uint256 price;
+        address token;
+        uint256 chainId;
+        address recipient;
+        uint256 destinationChainId;
+    }
+
```
fluffywaffles, line 123: Nit: extra newline.

src/builder/BridgeRoutes.sol
```diff
@@ -0,0 +1,465 @@
...
+        CCTP
+    }
+
+    struct Bridge {
+        // Note: Cannot name these `address` nor `type` because those are both reserved keywords
+        address bridgeAddress;
+        BridgeType bridgeType;
+    }
+
+    // TODO: convert strings to lower case before comparing. maybe rename to `compareSymbols`
```
fluffywaffles, line 147: Nb. alternatively, `stringEq` and e.g. `i_stringEq` for _i_nsensitive. :shrug: 

typically, anyway, `compare` suggests we describe how these compare; e.g., they are equal, less, greater, more or less precise, etc. For example, `Decimal.compare(a, b) == :gt` etc.

Often in lower-level APIs, `compare` functions return an integer somehow indicating the magnitude of difference. One such case is sorting algorithms.

I'd say `equal` is more appropriate.

src/builder/BridgeRoutes.sol
```diff
@@ -0,0 +1,465 @@
...
+    function compareStrings(string memory a, string memory b) internal pure returns (bool) {
+        return keccak256(abi.encodePacked((a))) == keccak256(abi.encodePacked((b)));
+    }
+
+    function filterAssetPositionsForSymbol(string memory assetSymbol, ChainAccount[] memory chainAccountsList) internal pure returns (AssetPositionsWithChainId[] memory) {
```
fluffywaffles, line 152: Nb. `ForSymbol` seems unnecessary here. Supposing we would support some other way of filtering, I'm not sure what it would be -- positions are already organized by chain ID in the input.

src/builder/BridgeRoutes.sol
```diff
@@ -0,0 +1,465 @@
...
+    }
+
+    struct AccountBalance {
+        address account;
+        uint256 balance;
+    }
+
+    struct Payment {
+        // TODO: rename?
+        bool isToken;
```
fluffywaffles, line 66: `payment.withToken`?

altho tbh I have no issues with `payment.isToken`.

src/builder/BridgeRoutes.sol
```diff
@@ -0,0 +1,465 @@
...
+            // Construct bridge operation if not enough funds on target chain
+            // TODO: bridge routing logic (which bridge to prioritize, how many bridges?)
+
+            // TODO: construct action contexts
+            if (payment.isToken) {
+                // wrap around paycall
+            } else {
+                address scriptAddress = getCodeAddress(codeJar, type(BridgeActions).creationCode);
+                bridgeQuarkOperation = QuarkOperation({
+                    nonce: , // TODO: get next nonce
```
fluffywaffles, line 368: this can be derived from `chainAccountsList[*].quarkStates`.

src/builder/BridgeRoutes.sol
```diff
@@ -0,0 +1,465 @@
...
+                    chainId: chainId,
```
fluffywaffles, line 369: seems wrong, no? this is tha destination chain ID.

The chain ID for the bridge operation would presumably be the source chain id.

Although I guess this entire code block is basically a placeholder, we don't really know how to pick a source chain right now anyway.

kevincheng96: Oops, my bad. This isn't supposed to be here. I forgot to save the file before committing it. Will push up the actual `BridgeRoutes` in a sec.

src/builder/BridgeRoutes.sol
kevincheng96: Yeah, I actually don't have a problem with it either. 👨‍🍳 was confused when I walked him through the code, so I left the TODO here to maybe rename it. `withToken` might be a bit more clear

src/builder/BridgeRoutes.sol
kevincheng96: Yeah, though this is block of code is not supposed to be in `BridgeRoutes` (see https://github.com/compound-finance/quark-scripts/pull/3#discussion_r1629880033). The latest commit no longer has this diff.

src/builder/BridgeRoutes.sol
kevincheng96: Could maybe filter based on `account` if we want to leave certain subaccounts untouched. But fair enough, can shorten the name for now.

src/builder/BridgeRoutes.sol
cwang25: Also we will need a map for chainId -> domainId (CCTP uses their own number for picking destination chain) for composing Bridge operation.
https://developers.circle.com/stablecoins/docs/supported-domains

src/builder/BridgeRoutes.sol
kevincheng96: Good suggestion. `stringEq` is good. Maybe `stringEqIgnoreCase` to capture the case insensitivity of the comparison.

src/builder/BridgeRoutes.sol
kevincheng96: Yeah. These constants are currently unused, so will probably be cleaned up in the PR that constructs Quark Operations and action contexts.

src/builder/BridgeRoutes.sol
kevincheng96: True. That'll be needed when constructing the Quark Operations (outside of scope for this PR).

src/builder/QuarkBuilder.sol
```diff
@@ -70,50 +122,93 @@ contract QuarkBuilder {
...
+                    index++;
+                }
+            }
+        }
+
+        return matchingAssetPositions;
     }
 
-    function getPriceFeed(string assetSymbol, uint256 chaindId) internal pure (address) {
+    function getAssetPositionForSymbolAndChain(string memory assetSymbol, uint256 chainId, ChainAccount[] memory chainAccountsList) internal pure returns (AssetPositionsWithChainId memory) {
```
fluffywaffles, line 175: Nit: there can be multiple positions for the same `{symbol, chain id}`. Multiple wallets.

src/builder/QuarkBuilder.sol
```diff
@@ -70,50 +122,93 @@ contract QuarkBuilder {
...
+                        accountBalances: assetPositions.accountBalances
+                    });
+                }
+            }
+        }
+
+        revert AssetPositionNotFound();
+    }
 
+    function sumUpBalances(AssetPositionsWithChainId memory assetPositions) internal pure returns (uint256) {
```
fluffywaffles, line 200: Nit: unnecessary "Up".

src/builder/QuarkBuilder.sol
```diff
@@ -122,39 +217,44 @@ contract QuarkBuilder {
...
         ChainAccounts[] calldata chainAccountsList
     ) external pure returns (BuilderResult memory) {
-        // TODO: Input validation: Check that arrays are equal length (e.g. chainIds and maxCosts in Payment)
-        AssetPositions[] paymentAssetAccounts;
-        // Get transfer and payment token
-        // TODO: implement filterAccounts
-        transferAssetAccounts = filterAccounts(assetSymbol, chainAccountsList);
+        if (payment.chainIds.length != payment.maxCosts.length) {
+            revert InvalidInput();
+        }
```
fluffywaffles: This will be obviated by refactoring the max costs data into an array-of-structs.

However, raises the question: if a chain id is not included in maxCosts, does that mean we should not consider assets on that chain? It is easily possible for the `chainAccountsList` to contain chains that have no specified max costs.

So, is that an error?
Or is that informative? Ie., if maxCosts is not empty, are those the only chains we should consider using?

src/builder/QuarkBuilder.sol
```diff
@@ -122,39 +217,44 @@ contract QuarkBuilder {
...
+            for (uint j = 0; j < transferAssetPositions[i].accountBalances.length; ++j) {
+                aggregateTransferAssetBalance += transferAssetPositions[i].accountBalances[j].balance;
             }
         }
-        if (transferAssetBalance < amount) {
+        if (aggregateTransferAssetBalance < amount) {
             revert InsufficientFunds();
         }
 
+        // TODO: Pay with bridged USDC?
```
fluffywaffles, line 236: Nit: "pay with bridged payment.currency?"

src/builder/QuarkBuilder.sol
```diff
@@ -122,39 +217,44 @@ contract QuarkBuilder {
...
         // MAX_COST_TOO_HIGH
-        // There are not enough funds to satisfy the total max payment cost, after transferring.
+        // There are not enough funds on each chain to satisfy the total max payment cost, after transferring.
```
fluffywaffles, line 238: Nit: `There is at least one chain that does not have sufficient payment assets to cover the maxCost for that chain.`

src/builder/BridgeRoutes.sol
kevincheng96: Done. Created a `PaymentMaxCost` struct.

src/builder/BridgeRoutes.sol
kevincheng96: Done

src/builder/BridgeRoutes.sol
kevincheng96: Done. Also moved these out into a separate `Strings` library.

src/builder/BridgeRoutes.sol
kevincheng96: Done

src/builder/BridgeRoutes.sol
kevincheng96: Left it as `isToken` for now. We can change it later if it proves to be too confusing.

src/builder/QuarkBuilder.sol
kevincheng96: Good catch, left out the `s` in `Positions`. Done

src/builder/QuarkBuilder.sol
kevincheng96: Done

src/builder/QuarkBuilder.sol
kevincheng96: Done

src/builder/QuarkBuilder.sol
kevincheng96: Much clearer, thanks

src/builder/QuarkBuilder.sol
kevincheng96: Yeah, this was removed after I added the `MaxPaymentCost` struct.

Hmm, interesting question. I don't think we should require the client to supply max costs for every chain. So, we could either 1) not construct Quark Operations on chains where max cost is unset or 2) choose some default value if max cost is unset. 

I do prefer 1, since there really isn't a good default value that works for all Quark Operations and payment tokens for 2. It's also a bit scary to set a value for it that's not specified by the client.

fluffywaffles: Agreed, I think if no max is provided, we should assume a max of 0.

fluffywaffles (approved):
lgtm on tests in a follow-up :+1:

src/builder/QuarkBuilder.sol
fluffywaffles: With the edge case that, if `maxCosts` is empty, we must still return a result that can be provided to simulate.

We may expect the client to explicitly pass in some large numbers for `maxCosts`, alternatively. :shrug: 

It's also perhaps a situation that will overlap with our handling of "maximum" or "unlimited" in the case of "transfer max" -- if we choose a sigil like `uint256(-1)` to represent "max" or "no limit," then maybe that's what we should use to explicitly construct simulations ignoring cost limits...

PR #10, opened 2024-06-10
Expiry

scott-silver:
Super basic, just adding a value for `expiry` when generating a `transferAsset` action.

The expiry value is the current block timestamp + a buffer. I've set it to an arbitrary value currently. We can bikeshed that actual value all we like.

### DIFF
--- a/src/builder/Actions.sol
+++ b/src/builder/Actions.sol
@@ -20,6 +20,7 @@ library Actions {
     string constant ACTION_TYPE_TRANSFER = "TRANSFER";
 
     uint256 constant TRANSFER_EXPIRY_BUFFER = 7 days;
+    uint256 constant BRIDGE_EXPIRY_BUFFER = 7 days;
 
     /* ===== Custom Errors ===== */
 
@@ -45,6 +46,7 @@ library Actions {
         address sender;
         uint256 destinationChainId;
         address recipient;
+        uint256 blockTimestamp;
     }
 
     /* ===== Output Types ===== */
@@ -110,7 +112,7 @@ library Actions {
                 bridge.originChainId, bridge.destinationChainId, bridge.amount, bridge.recipient, originUSDCPositions.asset
                 ),
             scriptSources: scriptSources,
-            expiry: 99999999999 // TODO: handle expiry
+            expiry: bridge.blockTimestamp + BRIDGE_EXPIRY_BUFFER
         });
     }
 
--- a/src/builder/Actions.sol
+++ b/src/builder/Actions.sol
@@ -19,7 +19,7 @@ library Actions {
     string constant ACTION_TYPE_BRIDGE = "BRIDGE";
     string constant ACTION_TYPE_TRANSFER = "TRANSFER";
 
-    uint256 constant TRANSFER_EXPIRY_BUFFER = 10_000;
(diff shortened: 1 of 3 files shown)


src/builder/Actions.sol
```diff
@@ -18,6 +18,8 @@ library Actions {
     string constant ACTION_TYPE_BRIDGE = "BRIDGE";
     string constant ACTION_TYPE_TRANSFER = "TRANSFER";
 
+    uint256 constant TRANSFER_EXPIRY_BUFFER = 10_000;
```
scott-silver, line 21: 10k is totally arbitrary

scott-silver: Hans suggested 7 days; we could update this to `7 * 24 * 60 * 60` or `7 days`

kevincheng96: Sure, 7 days makes sense.

src/builder/Actions.sol
```diff
@@ -84,12 +86,8 @@ library Actions {
...
-                bridge.originChainId,
-                bridge.destinationChainId,
-                bridge.amount,
-                bridge.recipient,
-                originUSDCPositions.asset
-            ),
+                bridge.originChainId, bridge.destinationChainId, bridge.amount, bridge.recipient, originUSDCPositions.asset
+                ),
             scriptSources: scriptSources,
             expiry: 99999999999 // TODO: handle expiry
```
kevincheng96, line 92: Is this one left out on purpose?

test/QuarkBuilder.t.sol
```diff
@@ -15,7 +15,7 @@ contract QuarkBuilderTest is Test {
         QuarkBuilder builder = new QuarkBuilder();
         vm.expectRevert(QuarkBuilder.InsufficientFunds.selector);
         builder.transfer(
-            transferUsdc_(1, 10_000_000, address(0xfe11a)), // transfer 1USDC on chain 1 to 0xfe11a
+            transferUsdc_(1, 10_000_000, address(0xfe11a), 123_456), // transfer 1USDC on chain 1 to 0xfe11a
```
kevincheng96, line 18: nit: Consider putting `123_456` in a constant (e.g. `BLOCK_TIMESTAMP`) to improve readability.

src/builder/Actions.sol
scott-silver: updated to use similar logic.

we're not testing this bridge action yet, yeah?

src/builder/QuarkBuilder.sol
```diff
@@ -116,7 +117,8 @@ contract QuarkBuilder {
                         sender: address(0), // FIXME: sender
                         // where it goes
                         destinationChainId: transferIntent.chainId,
-                        recipient: transferIntent.recipient
+                        recipient: transferIntent.recipient,
+                        blockTimestamp: transferIntent.blockTimestamp
```
scott-silver, line 121: this is incorrect, I believe. it should be the blockTimestamp for the origin chain ID?

src/builder/Actions.sol
kevincheng96: Nope we're not. There's still some work left to do on the implementation for bridge action.

kevincheng96 (approved):
👍

PR #12, opened 2024-06-12
Paycall Wrapper

cwang25:
Since I realize most codes in /main is now using Actions() that creates operation. 
So instead of making actions() to create paycall wrapped operation, I think it's better to create a designated library to wrap the resulting operation into paycall(). 
Which could gives flexibility on wrapping to another call (i.e. quotecall).


src/builder/PaycallWrapper.sol
```diff
@@ -0,0 +1,28 @@
...
+        internal
+        pure
+        returns (IQuarkWallet.QuarkOperation memory)
+    {
+        bytes[] memory scriptSources = new bytes[](1);
+        scriptSources[0] = type(Paycall).creationCode;
+
+        return IQuarkWallet.QuarkOperation({
+            nonce: operation.nonce,
+            scriptAddress: CodeJarHelper.getCodeAddress(chainId, type(Paycall).creationCode),
```
kevincheng96, line 20: The chain id shouldn't be needed after #13. CodeJar should have the same address on every chain.

src/builder/PaycallWrapper.sol
```diff
@@ -0,0 +1,28 @@
...
+
+// Helper library to wrap a QuarkOperation from Actions.sol for a Paycall
+library PaycallWrapper {
+    function wrap(IQuarkWallet.QuarkOperation memory operation, uint256 chainId, uint256 maxPaymentCost)
+        internal
+        pure
+        returns (IQuarkWallet.QuarkOperation memory)
+    {
+        bytes[] memory scriptSources = new bytes[](1);
+        scriptSources[0] = type(Paycall).creationCode;
```
kevincheng96, line 16: This should actually append the `Paycall` script source to the existing `scriptSources` in the `QuarkOperation`.

test/PaycallWrapper.t.sol
```diff
@@ -0,0 +1,62 @@
...
+
+    address constant USDC = 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48;
+
+    function setUp() public {
+        factory = new QuarkWalletProxyFactory(address(new QuarkWallet(new CodeJar(), new QuarkStateManager())));
+        codeJar = QuarkWallet(payable(factory.walletImplementation())).codeJar();
+    }
+
+    /* ===== general tests ===== */
+    function testSimpleTransferTransferAndWrapForPaycall() public {
```
kevincheng96, line 32: nit: duplicate `Transfer`?

kevincheng96 (commented):
Nice, looks clean!

src/builder/PaycallWrapper.sol
cwang25: I see. 
So it should be instead...
```
bytes[] memory scriptSources = new bytes[](operation.scriptSources.length + 1);
for(uint i = 0 ; i < operation.scriptSources.length ; i++){
    scriptSources[i] = operation.scriptSources[i];
}
scriptSources[scriptSources.length - 1] = type(Paycall).creationCode;
```
?

kevincheng96: Yeah that looks correct

test/PaycallWrapper.t.sol
cwang25: removed oops.

src/builder/PaycallWrapper.sol
cwang25: Fixed  :)

src/builder/PaycallWrapper.sol
cwang25: Sure will update the function call once it's merged. :)

src/builder/PaycallWrapper.sol
```diff
@@ -0,0 +1,31 @@
...
+    {
+        bytes[] memory scriptSources = new bytes[](operation.scriptSources.length + 1);
+        for (uint256 i = 0; i < operation.scriptSources.length; i++) {
+            scriptSources[i] = operation.scriptSources[i];
+        }
+        scriptSources[operation.scriptSources.length] = type(Paycall).creationCode;
+
+        return IQuarkWallet.QuarkOperation({
+            nonce: operation.nonce,
+            scriptAddress: CodeJarHelper.getCodeAddress(chainId, type(Paycall).creationCode),
```
kevincheng96, line 23: Paycall has a constructor, so you actually need to pass in something like:

`abi.encodePacked(type(Paycall).creationCode, abi.encode(ETH_USD_PRICE_FEED, USDC));`

kevincheng96: You may want to consider a library for PriceFeeds to get this information.

cwang25: Synced offline and created `PaymentToken.sol` to store the info we need on-chain for addresses.

src/builder/PaymentTokens.sol
```diff
@@ -0,0 +1,40 @@
...
+        paymentTokens[1] = PaymentToken({
+            chainId: 8453,
+            symbol: "USDC",
+            token: 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913,
+            priceFeed: 0x7e860098F58bBFC8648a4311b374B1D669a2bc6B
+        });
+        return paymentTokens;
+    }
+
+    function knownToken(uint256 chainId) internal pure returns (PaymentToken memory) {
```
kevincheng96, line 31: This should take the payment token as well right? There could be multiple payment tokens for one chain.

src/builder/PaycallWrapper.sol
```diff
@@ -12,15 +13,19 @@ library PaycallWrapper {
...
     {
+        PaymentTokens.PaymentToken memory paymentToken = PaymentTokens.knownToken(chainId);
+        bytes memory paycallSource =
+            abi.encodePacked(type(Paycall).creationCode, abi.encode(paymentToken.priceFeed, paymentToken.token));
         bytes[] memory scriptSources = new bytes[](operation.scriptSources.length + 1);
         for (uint256 i = 0; i < operation.scriptSources.length; i++) {
             scriptSources[i] = operation.scriptSources[i];
         }
-        scriptSources[operation.scriptSources.length] = type(Paycall).creationCode;
+
```
kevincheng96, line 23: nit: extra line

src/builder/PaymentTokens.sol
```diff
@@ -0,0 +1,40 @@
...
+            chainId: 1,
+            symbol: "USDC",
+            token: 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48,
+            priceFeed: 0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419
+        });
+        paymentTokens[1] = PaymentToken({
+            chainId: 8453,
+            symbol: "USDC",
+            token: 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913,
+            priceFeed: 0x7e860098F58bBFC8648a4311b374B1D669a2bc6B
```
kevincheng96, line 26: This seems like the wrong address. It's the address for USDC/USD

src/builder/PaymentTokens.sol
cwang25: Yeah,   was including it, but then realized we now only have USDC in here.  So wasn't sure if I should still include token symbol here to filter on. 
But I guess it's always better to make the function more generic for in future when adding more payment tokens for one chain will be easier.

cwang25: Just pushed another quick changes to included this function with tokenSymbol :) 
A

src/builder/PaymentTokens.sol
cwang25: oops fixed, and also added a doc strings to keep people in mind when adding more stuff in this file.

src/builder/PaycallWrapper.sol
cwang25: oh haha.  I'm used to this convention that always leave a extra line after loop `}` for readability 😄

src/builder/PaymentTokens.sol
```diff
@@ -17,21 +19,25 @@ library PaymentTokens {
...
             symbol: "USDC",
             token: 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913,
-            priceFeed: 0x7e860098F58bBFC8648a4311b374B1D669a2bc6B
+            priceFeed: 0x71041dddad3595F9CEd3DcCFBe3D1F4b0a16Bb70
         });
         return paymentTokens;
     }
 
-    function knownToken(uint256 chainId) internal pure returns (PaymentToken memory) {
+    function knownToken(string memory symbol, uint256 chainId) internal pure returns (PaymentToken memory) {
```
kevincheng96, line 37: nit: `tokenSymbol` or `assetSymbol`

kevincheng96 (approved):
Nice, LGTM 👍

src/builder/PaymentTokens.sol
cwang25: changed to tokenSymbol 👍

PR #16, opened 2024-06-13
Add paycall into builder code path

cwang25:
Add paycall wrapper into builder code path


src/builder/QuarkBuilder.sol
```diff
@@ -71,6 +75,12 @@ contract QuarkBuilder {
         uint256 blockTimestamp;
     }
 
+    struct BridgePayload {
+        uint256 amountLeftToBridge;
+        uint256 bridgeActionCount;
+        uint256 amountToBridge;
+    }
+
```
cwang25, line 83: By creating struct to hold just 3 more variables, we could avoid from having to use --via-ir.  
But downside is there is a struct that wrap holds multiple variables, might impact the readability of the code as people are used to initialize and declare variables by its own instead of a wrapper struct.

kevincheng96: I'd say let's not resort to this for now, since it does hurt readability. There also might be other ways to do this that are cleaner, like block scoping.

test/QuarkBuilder.t.sol
```diff
@@ -114,6 +118,69 @@ contract QuarkBuilderTest is Test {
...
+                })
+            ),
+            "action context encoded from TransferActionContext"
+        );
+        // TODO: actually generate digests
+        // assertNotEq0(result.quarkOperationDigest, hex"", "non-empty single digest");
+        // assertNotEq0(result.multiQuarkOperationDigest, hex"", "non-empty single digest");
+    }
+
+    // TODO: Test multiQuarkOperationDigest when Bridge operation logic is implemented
```
kevincheng96, line 183: This is already done in the test below

test/QuarkBuilder.t.sol
```diff
@@ -114,6 +118,69 @@ contract QuarkBuilderTest is Test {
...
+                    token: USDC_1,
+                    chainId: 1,
+                    recipient: address(0xceecee)
+                })
+            ),
+            "action context encoded from TransferActionContext"
+        );
+        // TODO: actually generate digests
+        // assertNotEq0(result.quarkOperationDigest, hex"", "non-empty single digest");
+        // assertNotEq0(result.multiQuarkOperationDigest, hex"", "non-empty single digest");
```
kevincheng96, line 180: Digests should be generated now, so these can be tested.

test/QuarkBuilder.t.sol
```diff
@@ -114,6 +118,69 @@ contract QuarkBuilderTest is Test {
         assertEq(result.multiQuarkOperationDigest, hex"", "empty multi digest");
     }
 
+    function testSimpleLocalTransferWithPaycallWrapperSucceeds() public {
```
kevincheng96, line 121: Would be good to test that the bridge operation is wrapped around a paycall successfully.

kevincheng96, line 121:
```suggestion
    function testSimpleLocalTransferWithPaycallSucceeds() public {
```

test/QuarkBuilder.t.sol
```diff
@@ -114,6 +118,69 @@ contract QuarkBuilderTest is Test {
...
+            result.quarkOperations[0].expiry, BLOCK_TIMESTAMP + 7 days, "expiry is current blockTimestamp + 7 days"
+        );
+        // check the actions
+        assertEq(result.actions.length, 1, "one action");
+        assertEq(result.actions[0].chainId, 1, "operation is on chainid 1");
+        assertEq(result.actions[0].quarkAccount, address(0xa11ce), "0xa11ce sends the funds");
+        assertEq(result.actions[0].actionType, "TRANSFER", "action type is 'TRANSFER'");
+        assertEq(result.actions[0].paymentMethod, "OFFCHAIN", "payment method is 'OFFCHAIN'");
+        assertEq(result.actions[0].paymentToken, address(0), "payment token is null");
+        assertEq(result.actions[0].paymentMaxCost, 0, "payment has no max cost, since 'OFFCHAIN'");
```
kevincheng96, line 164: These are no longer accurate. They should contain paycall information.

src/builder/QuarkBuilder.sol
```diff
@@ -312,4 +337,13 @@ contract QuarkBuilder {
...
         return result;
     }
+
+    function findMaxCost(Payment memory payment, uint256 chainId) internal pure returns (uint256) {
+        for (uint256 i = 0; i < payment.maxCosts.length; ++i) {
+            if (payment.maxCosts[i].chainId == chainId) {
+                return payment.maxCosts[i].amount;
+            }
+        }
+        return DEFAULT_MAX_PAYCALL_COST;
```
kevincheng96, line 347: I think the behavior we want is to actually ignore chains or throw an error if max cost is not provided for a chain.

test/QuarkBuilder.t.sol
cwang25: oh yeah, I think this is leftover code from old codebase when merging,will remove this comment.

src/builder/QuarkBuilder.sol
cwang25: Ok that sounds good 👍

src/builder/QuarkBuilder.sol
cwang25: Ok 👍

test/QuarkBuilder.t.sol
cwang25: Ok have updated to verify it as well. 👍

test/QuarkBuilder.t.sol
cwang25: Updated to have correct actions now. 
Now that I realized there is a separate file Actions.sol that creating those and those logics will closely related to the `payment.isToken` so I have move the switching logics into Actions.sol instead of handling it in QuarkBuilder.sol. 
For detecting when to wrap paycall.

scott-silver:
67 commits, but the total change is 300 lines? are all of those commits needed?

cwang25:
> 67 commits, but the total change is 300 lines? are all of those commits needed?

Yeah some back and forth, and this PR is old and was originally started on when the code was still "pseudo"  phase. 
But that's fine,  since we are going to "squash and merge" at the end, so in /main branch it won't be different than other PR merges.

src/builder/QuarkBuilder.sol
```diff
@@ -8,13 +8,18 @@ import {Accounts} from "./Accounts.sol";
...
 
 contract QuarkBuilder {
     /* ===== Constants ===== */
 
     string constant VERSION = "1.0.0";
     uint256 constant MAX_BRIDGE_ACTION = 1;
 
+    // Note: This is a default max cost for passing into paycall if PaymentMaxCost is missing for particular chainId
+    uint256 constant DEFAULT_MAX_PAYCALL_COST = 40e6;
+
```
kevincheng96, line 22: This can be removed

src/builder/QuarkBuilder.sol
```diff
@@ -261,28 +250,38 @@ contract QuarkBuilder {
     }
 
     // Assert that each chain has sufficient funds to cover the max cost for that chain.
-    // NOTE: This check assumes we will not be bridging payment tokens for the user.
+    // NOTE: This check strategy is:
+    // 1. Accrue available fund on each chain (chain balance - max cost)
+    // 2. If total available fund >= transferIntent then pass, otherwise MaxCostTooHigh()
+    // 3. Make sure the transferIntent chain always has enough fund to cover max cost
```
kevincheng96, line 256: I'm confused about what this change is for. It looks like you are no longer checking there is enough payment tokens to cover max payment costs on *all chains*, which seems wrong. Now, it only checks the target chain.

src/builder/QuarkBuilder.sol
```diff
@@ -164,22 +155,20 @@ contract QuarkBuilder {
...
+                chainAccountsList: chainAccountsList,
+                assetSymbol: transferIntent.assetSymbol,
+                amount: transferIntent.amount,
+                chainId: transferIntent.chainId,
+                sender: transferIntent.sender,
+                recipient: transferIntent.recipient,
+                blockTimestamp: transferIntent.blockTimestamp
+            }),
+            payment
+        );
```
kevincheng96, line 169: Nice, it looks cleaner to move the paycall wrapping to the `Actions` helper functions.

test/QuarkBuilder.t.sol
```diff
@@ -11,9 +11,16 @@ import {Actions} from "../src/builder/Actions.sol";
...
+import {Paycall} from "../src/Paycall.sol";
+import {PaycallWrapper} from "../src/builder/PaycallWrapper.sol";
+import {PaymentInfo} from "../src/builder/PaymentInfo.sol";
 
 contract QuarkBuilderTest is Test {
     uint256 constant BLOCK_TIMESTAMP = 123_456_789;
+    address constant ETH_USD_PRICE_FEED = 0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419;
+    address constant ETH_USD_PRICE_FEED_BASE = 0x71041dddad3595F9CEd3DcCFBe3D1F4b0a16Bb70;
+    address constant USDC_1 = address(0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48);
+    address constant USDC_8453 = address(0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913);
```
kevincheng96, line 23: You don't need to cast as `address()` here

kevincheng96, line 23: Let's be consistent with naming conventions. You're using `BASE` in one var and `8453` in another. I think let's stick with `8543` since it's already being used in some other places in this file.

test/QuarkBuilder.t.sol
```diff
@@ -11,9 +11,16 @@ import {Actions} from "../src/builder/Actions.sol";
 import {Accounts} from "../src/builder/Accounts.sol";
 import {CodeJarHelper} from "../src/builder/CodeJarHelper.sol";
 import {QuarkBuilder} from "../src/builder/QuarkBuilder.sol";
+import {Paycall} from "../src/Paycall.sol";
+import {PaycallWrapper} from "../src/builder/PaycallWrapper.sol";
+import {PaymentInfo} from "../src/builder/PaymentInfo.sol";
 
 contract QuarkBuilderTest is Test {
     uint256 constant BLOCK_TIMESTAMP = 123_456_789;
+    address constant ETH_USD_PRICE_FEED = 0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419;
```
kevincheng96, line 20: Post-pend with _1

test/QuarkBuilder.t.sol
```diff
@@ -308,6 +312,121 @@ contract QuarkBuilderTest is Test {
         assertNotEq(result.multiQuarkOperationDigest, hex"", "non-empty multi digest");
     }
 
+    function testSimpleBridgeTransferWithPaycallSucceeds() public {
+        QuarkBuilder builder = new QuarkBuilder();
+        PaymentInfo.PaymentMaxCost[] memory maxCosts = new PaymentInfo.PaymentMaxCost[](2);
+        maxCosts[0] = PaymentInfo.PaymentMaxCost({chainId: 1, amount: 1e5});
+        maxCosts[1] = PaymentInfo.PaymentMaxCost({chainId: 8453, amount: 1e5});
```
kevincheng96, line 319: nit: maybe use different amounts

kevincheng96 (commented):
Nice, looks better. I have a blocking question on `assertPaymentAffordable` since I don't think the new functionality looks correct.

src/builder/QuarkBuilder.sol
cwang25: Existing logics blindly add transfer.Intent on all the chains for payment need. 
So when user transfer actions that requires bridging, it will duplicated the amount required to transfer, and ended with `maxCostTooHigh()`

When updating the logics, I realized we should just check the chain that takes actions, as checking on all chains on maxCost might fail some actions that not necessary will fail. 

i.e. 
maxCost = $0.1 for Base
Mainnet = $3
Base = $0.05
And user wants to do transfer.  This check will fail when it checks user Base portfolio as $0.05 < $0.1

test/QuarkBuilder.t.sol
cwang25: I see. 
Yeah I kind of carried this over from the old dummy declaration of USDC_1  which was originally `address(0xaa)`

cwang25: Sure 👍

src/builder/QuarkBuilder.sol
cwang25: Synced offline, has adjusted wit the new assertPaymentAffordable logics.

src/builder/QuarkBuilder.sol
```diff
@@ -261,28 +262,53 @@ contract QuarkBuilder {
...
-        Payment memory payment
-    ) internal pure {
-        if (payment.isToken) {
-            for (uint256 i = 0; i < payment.maxCosts.length; ++i) {
-                uint256 paymentAssetBalanceOnChain = Accounts.sumBalances(
-                    Accounts.findAssetPositions(payment.currency, payment.maxCosts[i].chainId, chainAccountsList)
-                );
-                uint256 paymentAssetNeeded = payment.maxCosts[i].amount;
+    // Check user account can cover the cost of each actions
+    function assertActionsAffordable(Actions.Action[] memory actions, Accounts.ChainAccounts[] memory chainAccountsList)
```
cwang25, line 266: Now the affordable is check actions can be covered by the payment maxCost at the end of the build.

src/builder/QuarkBuilder.sol
```diff
@@ -261,28 +262,53 @@ contract QuarkBuilder {
...
+                amountUsed = transferActionContext.amount;
+            } else if (Strings.stringEqIgnoreCase(actions[i].actionType, Actions.ACTION_TYPE_BRIDGE)) {
+                Actions.BridgeActionContext memory bridgeActionContext =
+                    abi.decode(actions[i].actionContext, (Actions.BridgeActionContext));
+                tokenUsed = bridgeActionContext.token;
+                amountUsed = bridgeActionContext.amount;
+                plannedBridgeAmount += amountUsed;
+            } else {
+                revert InvalidActionType();
+            }
```
cwang25, line 287: In order to proper know the affordability of the actions when involve bridging + transferring actions on payment token. Will need to add `plannedBridgeAmount` when checking if target chain has enough fund.  
So that when check if target chain has enough fund to cover 5USDC won't fail when target chain only has 3USDC and 2USDC incoming from Base.

src/builder/QuarkBuilder.sol
```diff
@@ -99,6 +89,10 @@ contract QuarkBuilder {
             uint256 balanceOnDstChain =
                 Accounts.getBalanceOnChain(transferIntent.assetSymbol, transferIntent.chainId, chainAccountsList);
             uint256 amountLeftToBridge = transferIntent.amount - balanceOnDstChain;
+            // If the payment token is the transfer token and user opt for paying with the payment token, need to add max cost back to the amountLeftToBridge for target chain
+            if (payment.isToken && Strings.stringEqIgnoreCase(payment.currency, transferIntent.assetSymbol)) {
+                amountLeftToBridge += PaymentInfo.findMaxCost(payment, transferIntent.chainId);
+            }
```
cwang25, line 95: Since when we see `needsBridgedFunds()`, meaning the target chain is having less than the amount request on transfer. So here we check again and see if the transferring token is indeed payment token and opt for pay with token.  If it's true, then we need to add the maxCost to the balanceLeftToBridge as the target chain's balance will be all be used to transfer when computing the bridging logics in `balanceOnDstChain`

src/builder/QuarkBuilder.sol
```diff
@@ -242,7 +237,13 @@ contract QuarkBuilder {
...
                         )
                 ) {
-                    aggregateTransferAssetAvailableBalance += Accounts.sumBalances(positions);
+                    // If the payment token is the transfer token and user opt for paying with the payment token, reduce the available balance by the maxCost
+                    if (payment.isToken && Strings.stringEqIgnoreCase(payment.currency, transferIntent.assetSymbol)) {
+                        uint256 maxCost = PaymentInfo.findMaxCost(payment, chainAccountsList[i].chainId);
+                        aggregateTransferAssetAvailableBalance += Accounts.sumBalances(positions) - maxCost;
+                    } else {
+                        aggregateTransferAssetAvailableBalance += Accounts.sumBalances(positions);
+                    }
```
cwang25, line 246: For available fund when any actions (all actions will cost gas) to be applied onto it when token is payment token and user opt to pay with token.. Will always be balance - maxCost.  
So if we know user has 2USDC mainner (0.1 maxCost), and 2USDC on Base (0.01maxCost) and user request to transfer 3.99 USDC.  We can simply check (2 + 2 - 0.1 - 0.01 = 3.89) and failed user request, as we are sure with actions apply on to the quarkbuilder even with bridging,  it won't be enough.

remappings.txt
```diff
@@ -1,14 +1,15 @@
-test=test/
```
kevincheng96, line 1: This shouldn't be removed right? There are files in `test/` that need to be mapped here.

remappings.txt
```diff
@@ -1,14 +1,15 @@
...
-v3-periphery/=lib/v3-periphery/contracts/
-codejar/=lib/quark/src/codejar/
+quark-core-scripts/=lib/quark/src/quark-core-scripts/
 quark-core/=lib/quark/src/quark-core/
 quark-factory/=lib/quark/src/quark-factory/
 quark-proxy/=lib/quark/src/quark-proxy/
-src/=src/
\ No newline at end of file
+quark/=lib/quark/
+test/=lib/quark/test/
```
kevincheng96, line 13: I think this should be removed. We should not be referencing test files from the quark submodule.

remappings.txt
```diff
@@ -1,14 +1,15 @@
-test=test/
-@uniswap/=lib/
+@uniswap/=lib/quark/lib/
```
kevincheng96, line 1: Can this be removed? I don't see us using this remapping in the codebase.

src/builder/QuarkBuilder.sol
```diff
@@ -231,7 +237,13 @@ contract QuarkBuilder {
...
```
kevincheng96, line 246:
```suggestion
                    aggregateTransferAssetAvailableBalance += Accounts.sumBalances(positions);
                    // If the payment token is the transfer token and user opt for paying with the payment token, reduce the available balance by the maxCost
                    if (payment.isToken && Strings.stringEqIgnoreCase(payment.currency, transferIntent.assetSymbol)) {
                        uint256 maxCost = PaymentInfo.findMaxCost(payment, chainAccountsList[i].chainId);
                        aggregateTransferAssetAvailableBalance -= maxCost;
                    }
```

kevincheng96: nit suggestion: we don't need the `else` block if you keep L234 (`aggregateTransferAssetAvailableBalance += Accounts.sumBalances(positions);`). Feels cleaner this way to me personally.

src/builder/Strings.sol
```diff
@@ -7,7 +7,10 @@ library Strings {
     }
 
     function stringEqIgnoreCase(string memory a, string memory b) internal pure returns (bool) {
-        return keccak256(abi.encodePacked(toLowerCase(a))) == keccak256(abi.encodePacked(toLowerCase(b)));
+        // Need to copy bytes here to prevent unintentional memory changes side effect in strings that are passed in
+        string memory copyA = string(abi.encodePacked(a));
+        string memory copyB = string(abi.encodePacked(b));
```
kevincheng96, line 12: I don't understand why this is needed. Neither this function nor `toLowerCase` modifies `a` or `b`.

src/builder/QuarkBuilder.sol
```diff
@@ -250,38 +262,53 @@ contract QuarkBuilder {
...
+                plannedBridgeAmount += amountUsed;
+            } else {
+                revert InvalidActionType();
+            }
 
-                if (payment.maxCosts[i].chainId == transferIntent.chainId) {
-                    if (paymentAssetBalanceOnChain < paymentAssetNeeded) {
+            if (tokenUsed == actions[i].paymentToken) {
+                // If the payment token is the transfer token and this is the
+                // target chain, we need to account for the transfer amount
```
kevincheng96, line 291: nit: inconsistent line width

src/builder/QuarkBuilder.sol
```diff
@@ -250,38 +262,53 @@ contract QuarkBuilder {
...
-                uint256 paymentAssetNeeded = payment.maxCosts[i].amount;
-                if (paymentAssetBalanceOnChain > paymentAssetNeeded) {
-                    totalAvailableAsset += paymentAssetBalanceOnChain - paymentAssetNeeded;
-                }
+    // Check user account can cover the cost of each actions
+    function assertActionsAffordable(Actions.Action[] memory actions, Accounts.ChainAccounts[] memory chainAccountsList)
+        internal
+        pure
+    {
+        uint256 plannedBridgeAmount = 0;
```
kevincheng96, line 270: This needs to be 2-pass instead of 1-pass right? Without assuming ordering of the actions, you should be summing up `plannedBridgeAmount` first, and then running the validation logic afterwards.

src/builder/QuarkBuilder.sol
kevincheng96: Nice, makes sense to me 👍

src/builder/QuarkBuilder.sol
```diff
@@ -250,38 +262,53 @@ contract QuarkBuilder {
...
-                if (payment.maxCosts[i].chainId == transferIntent.chainId) {
-                    if (paymentAssetBalanceOnChain < paymentAssetNeeded) {
+            if (tokenUsed == actions[i].paymentToken) {
+                // If the payment token is the transfer token and this is the
+                // target chain, we need to account for the transfer amount
+                uint256 paymentAssetBalanceOnChain =
+                    Accounts.sumBalances(Accounts.findAssetPositions(tokenUsed, actions[i].chainId, chainAccountsList));
+
+                // If its transfer step, check if user has enough balance to cover the transfer amount after bridge
+                if (Strings.stringEqIgnoreCase(actions[i].actionType, Actions.ACTION_TYPE_TRANSFER)) {
```
kevincheng96, line 296: A cleaner way to write this instead of having multiple `if (actionType == Actions.ACTION_TYPE_X)` throughout the function is to first run a filter function to generate separate lists of actions for each type.

e.g. `Actions.Action[] memory bridgeActions = findBridgeActions(actions);`

Then, you can loop through each action type and process them separately.

test/QuarkBuilder.t.sol
```diff
@@ -315,7 +315,7 @@ contract QuarkBuilderTest is Test {
     function testSimpleBridgeTransferWithPaycallSucceeds() public {
         QuarkBuilder builder = new QuarkBuilder();
         PaymentInfo.PaymentMaxCost[] memory maxCosts = new PaymentInfo.PaymentMaxCost[](2);
-        maxCosts[0] = PaymentInfo.PaymentMaxCost({chainId: 1, amount: 1e5});
+        maxCosts[0] = PaymentInfo.PaymentMaxCost({chainId: 1, amount: 5e5});
         maxCosts[1] = PaymentInfo.PaymentMaxCost({chainId: 8453, amount: 1e5});
```
kevincheng96, line 319: Using `e5` is kind of confusing. These should be 0.5e6 and 0.1e6 right? Or just make them 5e6 and 1e6.

test/QuarkBuilder.t.sol
```diff
@@ -349,14 +349,14 @@ contract QuarkBuilderTest is Test {
                 abi.encodeWithSelector(
                     CCTPBridgeActions.bridgeUSDC.selector,
                     address(0xBd3fa81B58Ba92a82136038B25aDec7066af3155),
-                    2e6,
+                    21e5,
```
kevincheng96, line 352: Same here and below. Let's stick to `e6`. So `2.1e6` in this case.

src/builder/QuarkBuilder.sol
cwang25: Sounds good.👍

remappings.txt
cwang25: Hmm  ok... yeah that's weird, I didn't mean to update the remappings. 
I even purposedly remove the changes of my local rempappings.txt before pushing haha.
Ok let me re-run the forge remappings to regenerate this file.

cwang25: oof... looks like I did accidentally pushed the changes to remappings.txt
But now question is... I think that's what `forge remappings` returns.  So... should we keep it as is as the `forge remappings`?  

Cuz from my understanding `remappings.txt` doesn't do anything, but it's just to tell IDE to read.  And the true remappings is processed within `forge remappings` 🤔

src/builder/Strings.sol
cwang25: Actually it does which surprised me too. 

When I called this at the end of the generation (on the resulting action_context), the test failed and I realized all types failed with 'tranfser' != 'TRANFER' and 'BRDIGE_ACTION' != 'bridge_action' and realized it's was here that downcasing all the bytes.

But with thise copy and make string to copy with values,  the test will pass.

cwang25: To be clear, it's actually the `toLowerCase()` that changes the bytes. 
And when I was debating to add this copy value call between toLowerCase() or stringEqIgnoreCase(). 
I decided to put it here, as `stringEgIgnoreCase()` should always not changing the data from input.  
But `toLowerCase()` function, some might think they can just call that to change their string into lowercase.

cwang25: Another way works too as putting copy values into `toLowerCase()`

src/builder/QuarkBuilder.sol
cwang25: It's not necessary needs to be 2-pass.. I think 🤔 
Cuz the order still crucial here as user should follow the order of bridging first then transfer.  If user transfer first before bridging it should fail.  Unless we say that the order of QuarkBuilder is not guaranteed to be correct, and expect backend or client to proactively fix the order of the actions 🤔 (not likely?) In that case we will need 2-pass to account the case we get `bridge`-> `transfer` -> `bridge`  and expect BE or Client to find the third `bridge` and know to shift it to the front.

As for our current use case we must have 'bridge' actions before 'transfer' and only one 'transfer' at the end of the actions. So the `plannedBrdigeAmount` should have accrued all bridging amount before we're reaching 'transfer'

remappings.txt
cwang25: Reverted it to be matched with /main's  as I don't expect this PR should update any remappings entry. 
Apprently the `forge remappings` generate some wrongful data her.e

src/builder/QuarkBuilder.sol
cwang25: Changed to single line.

src/builder/QuarkBuilder.sol
kevincheng96: Yeah, the current code works because bridge actions are ordered in front of transfer actions. That's what I meant by saying "Without assuming ordering of the actions". The problem IMO is the code implicitly makes this assumption on ordering, which is prone to bugs and hurts readability. I would recommend either adjusting to 2-pass or calling out this assumption using comments.

src/builder/Strings.sol
kevincheng96: I don't think that's true. `toLowerCase()` shouldn't be modifying the input string, unless I'm misunderstanding something.

kevincheng96: Hmm...I just tested this locally and it is indeed modifying the input 🤔

cwang25: Yeah I was sort of surprised too. 
But with test failing with `TRANFER != transfer` and adding this the test passed right away sort of pointing to here as the culprit. 

I guess 
```
 bytes memory strBytes = bytes(str);
```
this is probably copy using the same reference as the position where it stores the string.

kevincheng96: Right, I think it makes sense to do it the way you have then. `toLowerCase` implies modifying the string in-place, so we can leave that as-is

src/builder/QuarkBuilder.sol
cwang25: Ok I have pushed out a newer version for this. 
So we verify two actions separately. 
And while processing bridgeAction we keep track on palnnedBridgeAmount. 

And when processing through the transactionAction, to have better future proof, we also track the accruedCost for multiple actions cases, instead of assuming it's just one action all the time.

src/builder/QuarkBuilder.sol
```diff
@@ -266,53 +266,93 @@ contract QuarkBuilder {
...
 
-                if (paymentAssetBalanceOnChain < actions[i].paymentMaxCost) {
+                if (paymentAssetBalanceOnChain < accruedCost) {
                     revert MaxCostTooHigh();
                 }
             }
         }
     }
 
+    function findActionsOfType(Actions.Action[] memory actions, string memory actionType)
```
kevincheng96, line 340: This should live in the Actions library. I guess `truncate` for Actions should also be moved there.

src/builder/QuarkBuilder.sol
```diff
@@ -266,53 +266,93 @@ contract QuarkBuilder {
...
-                    abi.decode(actions[i].actionContext, (Actions.TransferActionContext));
-                tokenUsed = transferActionContext.token;
-                amountUsed = transferActionContext.amount;
-            } else if (Strings.stringEqIgnoreCase(actions[i].actionType, Actions.ACTION_TYPE_BRIDGE)) {
-                Actions.BridgeActionContext memory bridgeActionContext =
-                    abi.decode(actions[i].actionContext, (Actions.BridgeActionContext));
-                tokenUsed = bridgeActionContext.token;
-                amountUsed = bridgeActionContext.amount;
-                plannedBridgeAmount += amountUsed;
+        // Verify bridge actions is affordable, and update plannedBridgeAmount for verifying transfer actions
```
kevincheng96, line 273: nit: "are affordable"

src/builder/QuarkBuilder.sol
```diff
@@ -266,53 +266,93 @@ contract QuarkBuilder {
...
+                if (paymentAssetBalanceOnChain < bridgeActions[i].paymentMaxCost) {
+                    revert MaxCostTooHigh();
+                }
             }
 
-            if (tokenUsed == actions[i].paymentToken) {
+            plannedBridgeAmount += bridgeActionContext.amount;
+        }
+
+        // Verify transfer actions is affordable
```
kevincheng96, line 304: nit: "are affordable"

src/builder/QuarkBuilder.sol
```diff
@@ -266,53 +266,93 @@ contract QuarkBuilder {
...
+                    revert MaxCostTooHigh();
+                }
             } else {
-                revert InvalidActionType();
+                // Just check payment token can cover the max cost
+                uint256 paymentAssetBalanceOnChain = Accounts.sumBalances(
+                    Accounts.findAssetPositions(
+                        bridgeActions[i].paymentToken, bridgeActions[i].chainId, chainAccountsList
+                    )
+                );
```
kevincheng96, line 294: This can be pulled out of the if statement and shared across the two cases.

src/builder/QuarkBuilder.sol
```diff
@@ -266,53 +266,93 @@ contract QuarkBuilder {
...
+            plannedBridgeAmount += bridgeActionContext.amount;
+        }
+
+        // Verify transfer actions is affordable
+        uint256 accruedCost = 0;
+        for (uint256 i = 0; i < transferActions.length; ++i) {
+            Actions.TransferActionContext memory transferActionContext =
+                abi.decode(transferActions[i].actionContext, (Actions.TransferActionContext));
+            address tokenUsed = transferActionContext.token;
+            uint256 amountUsed = transferActionContext.amount;
```
kevincheng96, line 310: nit: `transferToken`, `transferAmount`

src/builder/QuarkBuilder.sol
```diff
@@ -266,53 +266,93 @@ contract QuarkBuilder {
...
+            accruedCost += transferActions[i].paymentMaxCost;
+            if (tokenUsed == transferActions[i].paymentToken) {
                 // If the payment token is the transfer token and this is the target chain, we need to account for the transfer amount
-                uint256 paymentAssetBalanceOnChain =
-                    Accounts.sumBalances(Accounts.findAssetPositions(tokenUsed, actions[i].chainId, chainAccountsList));
+                uint256 paymentAssetBalanceOnChain = Accounts.sumBalances(
+                    Accounts.findAssetPositions(tokenUsed, transferActions[i].chainId, chainAccountsList)
+                );
```
kevincheng96, line 316: Same here, this can be pulled out and shared across the two cases.

src/builder/QuarkBuilder.sol
```diff
@@ -266,53 +266,93 @@ contract QuarkBuilder {
...
+                    revert MaxCostTooHigh();
+                }
             }
 
-            if (tokenUsed == actions[i].paymentToken) {
+            plannedBridgeAmount += bridgeActionContext.amount;
+        }
+
+        // Verify transfer actions is affordable
+        uint256 accruedCost = 0;
```
kevincheng96, line 305: Hmm, I think this really needs to be "accrued cost on each chain" right? It wouldn't work if the transfers were on different chains. To reduce complexity for now, I think it's safe to assume there's only one transfer and remove this variable.

cwang25: Yeah was thinking to use mapping,  but we can't use mapping which complicates thing here, and ended up with one variable haha. Since we only have one transferIntent.chainId,  so I ended up with assuming the transfer(s) will be on one target chain.

src/builder/QuarkBuilder.sol
cwang25: Sure I can move those existing functions to there as well.

src/builder/QuarkBuilder.sol
cwang25: ok 👍

src/builder/QuarkBuilder.sol
cwang25: Now I have changed it to have capability to support multi-chain transfers.

src/builder/QuarkBuilder.sol
```diff
@@ -262,37 +262,30 @@ contract QuarkBuilder {
...
 
         uint256 plannedBridgeAmount = 0;
-        // Verify bridge actions is affordable, and update plannedBridgeAmount for verifying transfer actions
+        // Verify bridge actions are affordable, and update plannedBridgeAmount for verifying transfer actions
         for (uint256 i = 0; i < bridgeActions.length; ++i) {
             Actions.BridgeActionContext memory bridgeActionContext =
                 abi.decode(bridgeActions[i].actionContext, (Actions.BridgeActionContext));
-
+            uint256 paymentAssetBalanceOnChain = Accounts.sumBalances(
+                Accounts.findAssetPositions(bridgeActionContext.token, bridgeActions[i].chainId, chainAccountsList)
```
kevincheng96, line 279: This has to be `bridgeActions[i].paymentToken`, not `bridgeActionContext.token` right? Would be good to have some tests because these bugs slipping through is not great :/

I suggest writing the test in a follow-up PR since this one is getting large and would be great to have it merged soon.

src/builder/QuarkBuilder.sol
```diff
@@ -301,79 +294,49 @@ contract QuarkBuilder {
...
+        );
+    }
+
+    function assertTransferActionsAffordableOnTargetChain(
+        Actions.Action[] memory transferActions,
+        Accounts.ChainAccounts[] memory chainAccountsList,
+        uint256 targetChainId,
+        uint256 plannedBridgeAmountToTargetChain
+    ) internal pure {
         uint256 accruedCost = 0;
```
kevincheng96, line 312: nit: `transferTokensUsed` is clearer

src/builder/QuarkBuilder.sol
cwang25: yeah. as for now all our test suites are based on USDC. 
Fxied.

src/builder/QuarkBuilder.sol
cwang25: 🤔 
But cost is not necessary to be transfertokenused?

kevincheng96: oops, I meant `paymentTokensUsed`

kevincheng96: Not blocking, we can always refactor in a future PR

kevincheng96 (approved):

src/builder/QuarkBuilder.sol
cwang25: done 👍

PR #17, opened 2024-06-13
Implement simple bridge algo

kevincheng96:
This implements a basic algo for figuring the amount to bridge from each chain. It loops through account balances on each chain and greedily attempts to bridge until the total amount needed is reached.

### DIFF
--- a/src/builder/Accounts.sol
+++ b/src/builder/Accounts.sol
@@ -88,4 +88,13 @@ library Accounts {
         }
         return totalBalance;
     }
+
+    function getBalanceOnChain(string memory assetSymbol, uint256 chainId, ChainAccounts[] memory chainAccountsList)
+        internal
+        pure
+        returns (uint256)
+    {
+        AssetPositions memory positions = findAssetPositions(assetSymbol, chainId, chainAccountsList);
+        return sumBalances(positions);
+    }
 }
--- a/src/builder/Actions.sol
+++ b/src/builder/Actions.sol
@@ -24,6 +24,7 @@ library Actions {
 
     /* ===== Custom Errors ===== */
 
+    error BridgingUnsupportedForAsset();
     error InvalidAssetForBridge();
 
     /* ===== Input Types ===== */
@@ -38,7 +39,7 @@ library Actions {
         uint256 blockTimestamp;
     }
 
-    struct BridgeUSDC {
+    struct BridgeAsset {
         Accounts.ChainAccounts[] chainAccountsList;
         string assetSymbol;
         uint256 amount;
@@ -83,7 +84,19 @@ library Actions {
         uint256 destinationChainId;
     }
 
-    function bridgeUSDC(BridgeUSDC memory bridge)
+    function bridgeAsset(BridgeAsset memory bridge)
+        internal
+        pure
+        returns (IQuarkWallet.QuarkOperation memory, Action memory)
+    {
+        if (Strings.stringEqIgnoreCase(bridge.assetSymbol, "USDC")) {
+            return bridgeUSDC(bridge);
+        } else {
+            revert BridgingUnsupportedForAsset();
+        }
+    }
+
+    function bridgeUSDC(BridgeAsset memory bridge)
         internal
         pure
         returns (IQuarkWallet.QuarkOperation memory, Action memory)
@@ -119,7 +132,6 @@ library Actions {
             amount: bridge.amount,
             price: srcUSDCPositions.usdPrice,
             token: srcUSDCPositions.asset,
-            // TODO: srcChainId
             chainId: bridge.srcChainId,
             recipient: bridge.recipient,
             destinationChainId: bridge.destinationChainId
--- a/src/builder/Actions.sol
+++ b/src/builder/Actions.sol
@@ -42,7 +42,7 @@ library Actions {
         Accounts.ChainAccounts[] chainAccountsList;
         string assetSymbol;
         uint256 amount;
-        uint256 originChainId;
+        uint256 srcChainId;
         address sender;
         uint256 destinationChainId;
         address recipient;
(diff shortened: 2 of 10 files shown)


src/builder/Accounts.sol
```diff
@@ -88,4 +88,13 @@ library Accounts {
         }
         return totalBalance;
     }
+
+    function getBalanceOnChain(string memory assetSymbol, uint256 chainId, ChainAccounts[] memory chainAccountsList)
```
hayesgm, line 92: These functions are pretty neat.

src/builder/Actions.sol
```diff
@@ -86,34 +86,56 @@ library Actions {
...
+            destinationChainId: bridge.destinationChainId
+        });
+        Action memory action = Actions.Action({
+            chainId: bridge.srcChainId,
+            quarkAccount: bridge.sender,
+            actionType: ACTION_TYPE_BRIDGE,
+            actionContext: abi.encode(bridgeActionContext),
+            paymentMethod: PAYMENT_METHOD_OFFCHAIN,
+            // Null address for OFFCHAIN payment.
+            paymentToken: address(0),
```
hayesgm, line 134: These should be paycallable, right?

src/builder/QuarkBuilder.sol
```diff
@@ -93,42 +96,80 @@ contract QuarkBuilder {
...
+                    continue;
+                }
+
+                Accounts.AssetPositions memory srcAssetPositions =
+                    Accounts.findAssetPositions(transferIntent.assetSymbol, srcChainAccounts.assetPositionsList);
+                Accounts.AccountBalance[] memory srcAccountBalances = srcAssetPositions.accountBalances;
+                // TODO: Make logic smarter. Currently, this uses a greedy algorithm.
+                for (uint256 j = 0; j < srcAccountBalances.length; ++j) {
+                    // TODO: May need to factor in bridge fees for non-CCTP bridges
+                    if (srcAccountBalances[j].balance < BRIDGE_MINIMUM_AMOUNT_USDC) {
```
hayesgm, line 130: I'm not against minimums, but this is weird since this constant is in USDC but we're... not checking the asset as USDC? So we'd basically need one parameter per bridgeable-asset? Plus this means we might get weird behavior like I want to send $1 over the bridge and it says "no, you can't"

src/builder/QuarkBuilder.sol
```diff
@@ -13,6 +13,8 @@ contract QuarkBuilder {
     /* ===== Constants ===== */
 
     string constant VERSION = "1.0.0";
+    uint256 constant BRIDGE_MINIMUM_AMOUNT_USDC = 1_000_000;
+    uint256 constant MAX_BRIDGE_ACTION = 2;
```
hayesgm, line 17: This should be 1 for now, right?

src/builder/QuarkBuilder.sol
```diff
@@ -93,42 +96,80 @@ contract QuarkBuilder {
...
+                    uint256 amountToBridge = srcAccountBalances[j].balance >= amountLeftToBridge
+                        ? amountLeftToBridge
+                        : srcAccountBalances[j].balance;
+                    amountLeftToBridge -= amountToBridge;
+
+                    if (payment.isToken) {
+                        // TODO: wrap around paycall
+                    } else {
+                        // TODO: Should change to handle non-USDC bridges later on
+                        (quarkOperations[actionIndex], actions[actionIndex]) = Actions.bridgeUSDC(
```
hayesgm, line 147: Is this actually `bridgeUSDC` or is this `bridgeCctp`? E.g. if we expanded to two USDC bridges, would this function be accurately named?

src/builder/QuarkBuilder.sol
hayesgm: I think the better approach, generally, would be to try and consolidate. As in, if it takes two bridge actions to accomplish what one can do, always do the one first. One way to represent this would be to (for instance) try this without allowing split bridging, and then try it again if that's insufficient _with_ split bridging.

src/builder/QuarkBuilder.sol
```diff
@@ -93,42 +96,80 @@ contract QuarkBuilder {
...
+                    !BridgeRoutes.canBridge(srcChainAccounts.chainId, transferIntent.chainId, transferIntent.assetSymbol)
+                ) {
+                    continue;
+                }
+
+                Accounts.AssetPositions memory srcAssetPositions =
+                    Accounts.findAssetPositions(transferIntent.assetSymbol, srcChainAccounts.assetPositionsList);
+                Accounts.AccountBalance[] memory srcAccountBalances = srcAssetPositions.accountBalances;
+                // TODO: Make logic smarter. Currently, this uses a greedy algorithm.
+                for (uint256 j = 0; j < srcAccountBalances.length; ++j) {
```
hayesgm, line 128: We should really order the actions by bridge priority, right?

src/builder/Actions.sol
kevincheng96: Yeah. Hans has some PRs for the `Paycall` logic, in which the `paymentMethod` here will change.

src/builder/QuarkBuilder.sol
kevincheng96: Sure, 1 makes more sense for now.

src/builder/QuarkBuilder.sol
kevincheng96: Right, we can sort the source chains by bridge priority.

src/builder/QuarkBuilder.sol
kevincheng96: Correct, we would ultimately need a config of minimums for each bridgeable asset. I have removed the minimum for now to simplify the logic a bit and give us room to explore alternatives (e.g. the one you suggested). I added a TODO for this.

src/builder/QuarkBuilder.sol
kevincheng96: Done. I generalized this from `bridgeUSDC` to `bridgeAsset`. The logic for `Actions.bridgeAsset` now can handle any type of asset. For unsupported assets (e.g. non-USDC as of now), the function will revert.

cwang25 (commented):
LGTM, 
once it's merge, I can help wrap the action into paycall/quotecall

cwang25 (approved):
LGTM, 
once it's merge, I can help wrap the action into paycall/quotecall

PR #18, opened 2024-06-13
Transfermax on quotecall

cwang25:
- Transfermax handling when user pass in `type(uint256).max`
- ~~Required quotecall, so it will be based on `hans/quotecall` branch~~ (merged)


src/builder/QuarkBuilder.sol
```diff
@@ -65,6 +66,13 @@ contract QuarkBuilder {
         Accounts.ChainAccounts[] memory chainAccountsList,
         PaymentInfo.Payment memory payment
     ) external pure returns (BuilderResult memory) {
+        // TransferMax flag
+        bool transferMax = transferIntent.amount == type(uint256).max;
+        // Convert transferIntent to user aggregated balance
+        if (transferMax) {
+            transferIntent.amount = totalAvailableAsset(transferIntent.assetSymbol, chainAccountsList, payment);
+        }
```
cwang25, line 74: Adjust transferIntent amount when seeing `type(uint256).max`

src/builder/QuarkBuilder.sol
```diff
@@ -120,9 +128,28 @@ contract QuarkBuilder {
...
+                                - PaymentInfo.findMaxCost(payment, srcChainAccounts.chainId);
+                        }
+                    } else {
+                        // Apply straighforward logics to bridge the token right away
+                        if (srcAccountBalances[j].balance >= amountLeftToBridge) {
+                            amountToBridge = amountLeftToBridge;
+                        } else {
+                            amountToBridge = srcAccountBalances[j].balance;
+                        }
+                    }
```
cwang25, line 151: Special handling + re-adjust for determining bridging amount when payment is payment. 
Because max cost is part of the balance of account's payment token.

test/QuarkBuilder.t.sol
```diff
@@ -435,6 +436,210 @@ contract QuarkBuilderTest is Test {
...
+            result.quarkOperations[1].scriptAddress,
+            quotecallAddressBase,
+            "script address[1] has been wrapped with quotecall address"
+        );
+        assertEq(
+            result.quarkOperations[1].scriptCalldata,
+            abi.encodeWithSelector(
+                Quotecall.run.selector,
+                CodeJarHelper.getCodeAddress(type(TransferActions).creationCode),
+                // Transfermax should be able to adjust to the right maxz amount to transfer all funds: 4e6 (holdings on base) - 0.1e6 (max cost on base) + 7.5e6 (brdiged from mainnet)= 3.9e6 + 7.5e6 = 11.4e6
```
kevincheng96, line 582: nit: typo in "max"

test/QuarkBuilder.t.sol
```diff
@@ -435,6 +436,210 @@ contract QuarkBuilderTest is Test {
...
+            "action context encoded from TransferActionContext"
+        );
+
+        // TODO: Check the contents of the EIP712 data
+        assertNotEq(result.eip712Data.digest, hex"", "non-empty digest");
+        assertNotEq(result.eip712Data.domainSeparator, hex"", "non-empty domain separator");
+        assertNotEq(result.eip712Data.hashStruct, hex"", "non-empty hashStruct");
+    }
+
+    function testSimpleBridgeTransferMax() public {
```
kevincheng96, line 513: Nice test!

src/builder/Actions.sol
```diff
@@ -154,7 +159,9 @@ library Actions {
             quarkAccount: bridge.sender,
             actionType: ACTION_TYPE_BRIDGE,
             actionContext: abi.encode(bridgeActionContext),
-            paymentMethod: payment.isToken ? PAYMENT_METHOD_PAYCALL : PAYMENT_METHOD_OFFCHAIN,
+            paymentMethod: payment.isToken
+                ? useQuoteCall ? PAYMENT_METHOD_QUOTECALL : PAYMENT_METHOD_PAYCALL
+                : PAYMENT_METHOD_OFFCHAIN,
```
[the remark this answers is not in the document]
cwang25: Sure 👍 Will make a if clauses for assigning this value.

I thought 3 is not too bad yet haha. 😄

src/builder/QuarkBuilder.sol
```diff
@@ -345,4 +374,25 @@ contract QuarkBuilder {
             }
         }
     }
+
+    function totalAvailableAsset(
```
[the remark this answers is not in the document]
cwang25: Sounds good. 

I think I left it here because I see `assertSuffientFunds` and `assertFundsAvailable`  Make me wonder if we should move all that to Accounts.sol as well.  🤔  (or strip off some account-only logics into Accounts.sol)

src/builder/QuarkBuilder.sol
```diff
@@ -345,4 +374,25 @@ contract QuarkBuilder {
...
+        for (uint256 i = 0; i < chainAccountsList.length; ++i) {
+            total += Accounts.sumBalances(
+                Accounts.findAssetPositions(tokenSymbol, chainAccountsList[i].chainId, chainAccountsList)
+            );
+
+            // Account for max cost if the payment token is the transfer token
+            // Simply offset the max cost from the available asset batch
+            if (payment.isToken && Strings.stringEqIgnoreCase(payment.currency, tokenSymbol)) {
+                total -= PaymentInfo.findMaxCost(payment, chainAccountsList[i].chainId);
+            }
```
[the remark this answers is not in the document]
cwang25: My thought was to try to make it not rely on TransferIntent as non bridgeable errors could be caught in `assertFundsAvailable` which will take in transferIntent and know which chain need to bridge or not and throw errors if needed. 

But this actually goes into a interesting UX case as user has 
```
User has: 
10 USDC on mainnet
5 USDC  on Base
4 USDC on Unbridgeable L2

User requests transfermax on mainnet. 
Do we: 
A)
Trying to transfermax on 19 USDC, then throw user errors with `FundsUnavailable()` from assertFundsAvailable

B)
We know 4USDC is unbridgable, so we adjust that so that even user requests to transfermax,  we only tranfser 15USDC and ignore 4USDC on Unbridgeable L2.  
After actions completes, user will see 15USDC transferrred out on mainnet,  but will still see 4USDC on Unbridgeable L2.
```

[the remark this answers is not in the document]
cwang25: Discussed offline in discord, 
Looks like we're going with A).

test/QuarkBuilder.t.sol
cwang25: Thanks 😄

src/builder/QuarkBuilder.sol
```diff
@@ -178,7 +186,11 @@ contract QuarkBuilder {
             }
 
             if (amountLeftToBridge > 0) {
-                revert FundsUnavailable();
+                revert FundsUnavailable(
+                    transferIntent.amount - balanceOnDstChain,
+                    transferIntent.amount - amountLeftToBridge,
+                    amountLeftToBridge
+                );
```
kevincheng96, line 193: Just a heads up I changed this custom error in my bridging logic refactor: https://github.com/compound-finance/quark-scripts/pull/30/files#diff-cd5318530a09b8df7f1d19dcd7fa22875ac1c8b9a51298e8f7b4e6824cf89b83R33

Nothing for you to do, but rather just letting you know how this might evolve.

src/builder/QuarkBuilder.sol
```diff
@@ -178,7 +186,11 @@ contract QuarkBuilder {
             }
 
             if (amountLeftToBridge > 0) {
-                revert FundsUnavailable();
+                revert FundsUnavailable(
+                    transferIntent.amount - balanceOnDstChain,
+                    transferIntent.amount - amountLeftToBridge,
```
kevincheng96, line 191: This should be `requiredAmount - amountLeftToBridge => transferIntent.amount - balanceOnDstChain - amountLeftToBridge` right?

src/builder/QuarkBuilder.sol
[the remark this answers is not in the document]
kevincheng96: Do we have a test for A? Correct me if I'm wrong, but I don't see any diffs related to the decision here.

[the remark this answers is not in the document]
cwang25: Oh yeah I don't expect logics diff occurs in for A,  as the follow-up assertFundAvailable will throw errors. 
But yeah I can add a test case to verify in A) case, we will get FundsUnavailable() error.

src/builder/QuarkBuilder.sol
cwang25: oops yeah 😅   forgot the DstChain that don't need to brdige ... oops.

src/builder/QuarkBuilder.sol
cwang25: Oh yeah I saw that too 😄 
I modified it to use the same exact naming as well. 

But I found it might be useful to leave in the `missingAmount`  so that client don't have to do `requestAmount - actualAmount` when they sees errors.

src/builder/QuarkBuilder.sol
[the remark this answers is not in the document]
cwang25: I have added a test case to cover for A.  👍

test/QuarkBuilder.t.sol
```diff
@@ -641,6 +643,42 @@ contract QuarkBuilderTest is Test {
...
+        chainAccountsList[2] = Accounts.ChainAccounts({
+            chainId: 7777,
+            quarkStates: quarkStates_(address(0xfe11a), 2),
+            assetPositionsList: assetPositionsList_(7777, address(0xfe11a), 5e6)
+        });
+
+        // User has total holding of 17 USDC, but only 12 USDC is available for transfer/bridge to 8453, and missing 5 USDC stuck in random L2 so will revert with FundsUnavailable error
+        // Actual number re-adjusted with max cost gas fee so that the:
+        // Request amount = 17 USDC - 0.5 USDC (max cost on main) - 0.1 USDC (max cost on Base) - 0.1 USDC (max cost on RandomL2) = 16.3 USDC
+        // Actual amount = 8 USDC (available on main) + 4 USDC (available on Base) - 0.5 USDC - 0.1 UDC = 11.4 USDC
```
kevincheng96, line 672: "USDC"

test/QuarkBuilder.t.sol
```diff
@@ -641,6 +643,42 @@ contract QuarkBuilderTest is Test {
...
+        });
+        chainAccountsList[2] = Accounts.ChainAccounts({
+            chainId: 7777,
+            quarkStates: quarkStates_(address(0xfe11a), 2),
+            assetPositionsList: assetPositionsList_(7777, address(0xfe11a), 5e6)
+        });
+
+        // User has total holding of 17 USDC, but only 12 USDC is available for transfer/bridge to 8453, and missing 5 USDC stuck in random L2 so will revert with FundsUnavailable error
+        // Actual number re-adjusted with max cost gas fee so that the:
+        // Request amount = 17 USDC - 0.5 USDC (max cost on main) - 0.1 USDC (max cost on Base) - 0.1 USDC (max cost on RandomL2) = 16.3 USDC
```
kevincheng96, line 671: Shouldn't this amount be `16.3 - 4 = 12.3` USDC? 4 USDC is already on chain 8453, meaning only 12.3 need to be bridged over.

Also, I think the math/formula might be wrong, as it double subtracts the max cost on the destination chain. e.g. 0.1 USDC will be subtracted twice. Once during the `totalAvailableAsset` calculation, and a second when subtracting the `balanceOnDstChain`.

cwang25: Hmm oh yeah, this is a bit confusing. 
But as this test case will actually get caught by asserFundsAvailable() way before the bridge actions take in place,  I'm still trying to see if I can find a way to "pass" `assertFundsAvailable()` but fail at bridge, since `assertFundsAvailable()` verify the `canBridge()` also. 

If based on the `assertFundsAvailable()` logics, then it is 16.3. 
As the request amount is the origin maxbalance, which is sum of all 3 chains - cost for each chain = 16.4

I think if we pass the `assertFundsAvailable()` and fail in the bridge actions, then yeah it should be 12.3.  
But I need to find the way to pass `assertFundsAvailable()` for that case. lol 😅 
-----------------------------

Re about forumula on `totalAvailableAsset` 
I'm trying to list out the calculation here,  I think I didn't see where it double substracted in balanceOnDstChain 😅 
(Maybe I'm missing something obvious here and probably need another 👀 to double check) 
```
User has A USDC on main, B USDC on Base, C USDC on Blast
Cost is a, b, c on main, base, blast. 
Transfermax on main.

So in bridge blocks variables looks like ... 

transferIntent.Amount = totalAvailableAsset = A + B + C - a- b - c
balanceOnSDstChain = Accounts.getBalanceOnChain(transferIntent.assetSymbol, transferIntent.chainId, chainAccountsList) = A

amountLeftToBridge = (A+B+C-a-b-c) - A = B+C-a-b-c
# Need to add `a` back to amountLeftToBridge to account for dst chain gas 
if (paymentToken is tranfer):
   amountLeftToBridge = B+C-a-b-c + a = B+C-b-c

```

cwang25: Part of me feel like `assertFundsAvailable()` will catch most errors that bridge actions will run into. 
If assertFundsAvailable() pass,  I think it might be safe to assume the bridge actions will pass successfully. 🤔

kevincheng96: Let me think about it a bit more, but I guess if we can't think of a test case that hits this error, that really means some of the code is unnecessary and should be removed right?

src/builder/QuarkBuilder.sol
```diff
@@ -128,9 +139,28 @@ contract QuarkBuilder {
...
+                            srcAccountBalances[j].balance
+                                >= amountLeftToBridge + PaymentInfo.findMaxCost(payment, srcChainAccounts.chainId)
+                        ) {
+                            amountToBridge = amountLeftToBridge;
+                        } else {
+                            amountToBridge = srcAccountBalances[j].balance
+                                - PaymentInfo.findMaxCost(payment, srcChainAccounts.chainId);
+                        }
+                    } else {
+                        // Apply straighforward logics to bridge the token right away
```
kevincheng96, line 156: nit: typo in straightforward

src/builder/QuarkBuilder.sol
```diff
@@ -128,9 +139,28 @@ contract QuarkBuilder {
...
+                    if (payment.isToken && Strings.stringEqIgnoreCase(payment.currency, transferIntent.assetSymbol)) {
+                        // Apply payment token logics to figure out the amount to bridge
+                        if (
+                            srcAccountBalances[j].balance
+                                >= amountLeftToBridge + PaymentInfo.findMaxCost(payment, srcChainAccounts.chainId)
+                        ) {
+                            amountToBridge = amountLeftToBridge;
+                        } else {
+                            amountToBridge = srcAccountBalances[j].balance
+                                - PaymentInfo.findMaxCost(payment, srcChainAccounts.chainId);
```
kevincheng96, line 153: I think this logic only makes sense if there is a single account on each chain. I think that's a fair assumption to make now, but let's leave a `NOTE:` detailing this assumption.

test/QuarkBuilder.t.sol
kevincheng96: Thinking about this again, I think the test case is fine actually. In my bridge refactor PR (#30), I switch out the untested `FundsAvailable` error for a more specific `NotEnoughFundsToBridge`, which I test there.

kevincheng96 (approved):

test/QuarkBuilder.t.sol
cwang25: Interesting, yeah #30 could reach that when there is different flow to bridge payment token. 

Yeah I think the assertFundsAvialable() only catches only on transfer token and when transfer token is also payment token.

PR #21, opened 2024-06-17
Update README

kevincheng96:
The README was forked from the `quark` repo, so this PR adapts it to be `quark-scripts`-specific. I also add a section to describe Quark Builder.


scott-silver (approved):
🙌

PR #22, opened 2024-06-20
Mark some wayward library functions as pure internal

coburncoburn:


scott-silver (approved):

src/builder/CodeJarHelper.sol
```diff
@@ -7,7 +7,7 @@ library CodeJarHelper {
     /// @notice The address for CodeJar on all chains
     address constant CODE_JAR_ADDRESS = 0x2b68764bCfE9fCD8d5a30a281F141f69b69Ae3C8;
 
-    function getCodeAddress(bytes memory code) public pure returns (address) {
+    function getCodeAddress(bytes memory code) pure internal returns (address) {
```
scott-silver, line 10: looks like this should be `internal pure returns`

kevincheng96 (approved):

PR #23, opened 2024-06-20
Remove unnecessary fields in structs

kevincheng96:
This PR removes unused fields from `QuarkState` and slims down `BuilderResult` to have a single digest instead of both `multiQuarkOperationDigest` and `quarkOperationDigest`.


coburncoburn (approved):
makes sense to me

PR #24, opened 2024-06-20
Add QuarkBuilderHelper

kevincheng96:
`QuarkBuilderHelper` is a contract of helper functions that might be useful for clients using the `QuarkBuilder`.

### DIFF
--- a/src/builder/QuarkBuilderHelper.sol
+++ b/src/builder/QuarkBuilderHelper.sol
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity ^0.8.23;
+
+import {BridgeRoutes} from "./BridgeRoutes.sol";
+
+/**
+ * @title Quark Builder Helper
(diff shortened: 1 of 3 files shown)


src/builder/QuarkBuilderHelper.sol
```diff
@@ -0,0 +1,27 @@
...
+        external
+        pure
+        returns (bool)
+    {
+        return BridgeRoutes.canBridge(srcChainId, dstChainId, assetSymbol);
+    }
+
+    function addBufferToPaymentCost(uint256 maxPaymentCost) external pure returns (uint256) {
+        return maxPaymentCost * PAYMENT_COST_BUFFER / BUFFER_SCALE;
+    }
```
cwang25, line 26: Looks like it's based on constant, and always just add 20% on top of the maxParymentCost? 🤔

cwang25 (approved):

src/builder/QuarkBuilderHelper.sol
kevincheng96: Correct, that's the heuristic we use for now.

PR #25, opened 2024-06-20
potential remapping fix

scott-silver:


test/QuarkBuilder.t.sol
```diff
@@ -12,6 +12,8 @@ import {Accounts} from "../src/builder/Accounts.sol";
 import {CodeJarHelper} from "../src/builder/CodeJarHelper.sol";
 import {QuarkBuilder} from "../src/builder/QuarkBuilder.sol";
 
+import "./NewFile.sol";
```
kevincheng96, line 15: Doesn't this need to be "test/NewFile.sol" to test the remapping?

scott-silver: I believe both work correctly; just pushed `test/NewFile.sol` to confirm

kevincheng96 (commented):
LGTM, seems like a bug that `test` was not postpended with `test/`. I see this convention being used in the `quark` repo `remappings.txt` as well, so we might want to fix it there too.

kevincheng96 (approved):

PR #26, opened 2024-06-21
Add domainSeparator and structHash to BuilderResult

kevincheng96:
We use a new `EIP712Data` struct to hold all the EIP-712 related data now, since there are three values we return (digest, domain separator, struct hash).


hayesgm (approved):
image

PR #27, opened 2024-06-24
add husks to get actionContext abis

coburncoburn:
The ABIs for these structs are things that clients depend upon, but they are not available from the compiled abi as they are serialized before being given as a return value. These simple functions serve to get their abis into the out/actions.sol/actions.json.

### DIFF
--- a/src/builder/Actions.sol
+++ b/src/builder/Actions.sol
@@ -268,4 +268,17 @@ library Actions {
         }
         return result;
     }
-}
+
+    // These structs are mostly used internally and returned in serialzed format as bytes: actionContext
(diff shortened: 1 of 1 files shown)


hayesgm (approved):

src/builder/Actions.sol
```diff
@@ -268,4 +268,17 @@ library Actions {
         }
         return result;
     }
-}
+
+    // These structs are mostly used internally and returned in serialzed format as bytes: actionContext
```
kevincheng96, line 272: nit: typo in serialized

kevincheng96 (approved):
You'll need to run the formatter before merging

PR #28, opened 2024-06-24
Ignore chain if no max cost

kevincheng96:
If the action is paid for with tokens and max cost is not specified for a chain, then the Builder will now ignore that chain instead of reverting. This allows clients to indicate the chains an action can utilize.

### DIFF
--- a/src/builder/QuarkBuilder.sol
+++ b/src/builder/QuarkBuilder.sol
@@ -21,7 +21,7 @@ contract QuarkBuilder {
 
     error AssetPositionNotFound();
     error FundsUnavailable();
-    error InsufficientFunds();
+    error InsufficientFunds(uint256 requiredAmount, uint256 actualAmount);
     error InvalidInput();
     error MaxCostTooHigh();
     error TooManyBridgeOperations();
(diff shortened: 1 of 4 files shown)


src/builder/QuarkBuilder.sol
```diff
@@ -218,40 +226,51 @@ contract QuarkBuilder {
...
+                if (payment.isToken && Strings.stringEqIgnoreCase(payment.currency, transferIntent.assetSymbol)) {
+                    uint256 maxCost = PaymentInfo.findMaxCost(payment, chainAccountsList[i].chainId);
+                    aggregateTransferAssetAvailableBalance -= maxCost;
+                }
             }
         }
+        if (aggregateTransferAssetAvailableBalance < transferIntent.amount) {
+            revert FundsUnavailable();
+        }
     }
```
kevincheng96, line 274: The diff here looks big, but most of the diff is just removing one layer of nesting by moving the outer check (`if needsBridgedFunds`) out. The real diff is adding this check:

```
// Skip the chain if action is paid for with token and max cost is not specified for the current chain
if (payment.isToken && !PaymentInfo.hasMaxCostForChain(payment, chainAccountsList[i].chainId)) {
    continue;
 }
```

cwang25:
wondering if ignoring the chain with no max cost, whether or not we could just start ignoring from the very top by filtering `chainAccountsList` so the ignoring chain could be "invisible" through the whole complex logics blocks following after. 🤔

kevincheng96:
> wondering if ignoring the chain with no max cost, whether or not we could just start ignoring from the very top by filtering `chainAccountsList` so the ignoring chain could be "invisible" through the whole complex logics blocks following after. 🤔

That's a good suggestion, I like it. It's quite a bit cleaner that way IMO.

kevincheng96:
@cwang25 Updated with your suggestion

cwang25 (approved):
LGTM 👍

PR #29, opened 2024-06-25
CometSupply

scott-silver:
~~Not quite done, but wanted to share in-progress version since I believe it's starting to become a blocker.~~

The CometSupply action is almost entirely adapted from the existing Tranfser action; I've tried to abstract shared functionality, but there are some places where it's somewhat awkward to share that functionality (specifically in the bridging logic).

For the test files, I've created a separate `QuarkBuilderTest` contract that our QuarkBuilder tests can inherit from. I moved the existing Transfer tests into QuarkBuilderTransfer.t.sol and wrote the new tests in QuarkBuilderCometSupply.t.sol.

PR #30, opened 2024-06-25
Bridge payment token if not enough

kevincheng96:
This PR implements logic to bridge the payment token if there is not enough on the destination chain. I also extract out the bridge operation construction logic into a helper function because it is used in two places right now and will be used in more in the future.

### DIFF
--- a/src/builder/Actions.sol
+++ b/src/builder/Actions.sol
@@ -91,6 +91,7 @@ library Actions {
         // Address of payment token on chainId.
         // Null address if the payment method was OFFCHAIN.
         address paymentToken;
+        string paymentTokenSymbol;
         uint256 paymentMaxCost;
     }
 
@@ -107,6 +108,7 @@ library Actions {
 
     struct BridgeActionContext {
         uint256 amount;
+        string assetSymbol;
         string bridgeType;
         uint256 chainId;
         uint256 destinationChainId;
@@ -153,6 +155,7 @@ library Actions {
 
     struct SupplyActionContext {
         uint256 amount;
+        string assetSymbol;
         uint256 chainId;
         address comet;
         uint256 price;
@@ -161,6 +164,7 @@ library Actions {
 
     struct TransferActionContext {
         uint256 amount;
+        string assetSymbol;
         uint256 chainId;
         uint256 price;
         address recipient;
@@ -245,6 +249,7 @@ library Actions {
             amount: bridge.amount,
             price: srcUSDCPositions.usdPrice,
             token: srcUSDCPositions.asset,
+            assetSymbol: srcUSDCPositions.symbol,
             chainId: bridge.srcChainId,
             recipient: bridge.recipient,
             destinationChainId: bridge.destinationChainId,
@@ -267,6 +272,7 @@ library Actions {
             paymentMethod: paymentMethod,
             // Null address for OFFCHAIN payment.
             paymentToken: payment.isToken ? PaymentInfo.knownToken(payment.currency, bridge.srcChainId).token : address(0),
+            paymentTokenSymbol: payment.currency,
             paymentMaxCost: payment.isToken ? PaymentInfo.findMaxCost(payment, bridge.srcChainId) : 0
         });
 
@@ -322,7 +328,8 @@ library Actions {
             chainId: cometSupply.chainId,
             comet: cometSupply.comet,
             price: assetPositions.usdPrice,
-            token: assetPositions.asset
+            token: assetPositions.asset,
+            assetSymbol: assetPositions.symbol
         });
         Action memory action = Actions.Action({
             chainId: cometSupply.chainId,
@@ -332,6 +339,7 @@ library Actions {
             paymentMethod: payment.isToken ? PAYMENT_METHOD_PAYCALL : PAYMENT_METHOD_OFFCHAIN,
             // Null address for OFFCHAIN payment.
             paymentToken: payment.isToken ? PaymentInfo.knownToken(payment.currency, cometSupply.chainId).token : address(0),
+            paymentTokenSymbol: payment.currency,
             paymentMaxCost: payment.isToken ? PaymentInfo.findMaxCost(payment, cometSupply.chainId) : 0
         });
 
@@ -391,6 +399,7 @@ library Actions {
             amount: transfer.amount,
             price: assetPositions.usdPrice,
             token: assetPositions.asset,
+            assetSymbol: assetPositions.symbol,
             chainId: transfer.chainId,
             recipient: transfer.recipient
         });
@@ -410,6 +419,7 @@ library Actions {
             paymentMethod: paymentMethod,
             // Null address for OFFCHAIN payment.
             paymentToken: payment.isToken ? PaymentInfo.knownToken(payment.currency, transfer.chainId).token : address(0),
+            paymentTokenSymbol: payment.currency,
             paymentMaxCost: payment.isToken ? PaymentInfo.findMaxCost(payment, transfer.chainId) : 0
         });
 
--- a/src/builder/Actions.sol
+++ b/src/builder/Actions.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 pragma solidity ^0.8.23;
 
-import {CCTP} from "./BridgeRoutes.sol";
+import {BridgeRoutes, CCTP} from "./BridgeRoutes.sol";
 import {Strings} from "./Strings.sol";
 import {Accounts} from "./Accounts.sol";
 import {CodeJarHelper} from "./CodeJarHelper.sol";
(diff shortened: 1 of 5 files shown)


src/builder/QuarkBuilder.sol
```diff
@@ -308,23 +295,33 @@ contract QuarkBuilder {
...
         );
     }
 
     function assertTransferActionsAffordableOnTargetChain(
         Actions.Action[] memory transferActions,
         Accounts.ChainAccounts[] memory chainAccountsList,
         uint256 targetChainId,
-        uint256 plannedBridgeAmountToTargetChain
+        uint256 transferTokenBridgeAmount,
+        uint256 paymentTokenBridgeAmount
```
kevincheng96, line 324: We actually need the amount bridged for both the transfer token and the payment token to do this check.

src/builder/Actions.sol
```diff
@@ -83,12 +94,96 @@ library Actions {
...
         uint256 price;
         address token;
+        string assetSymbol;
         uint256 chainId;
         address recipient;
         uint256 destinationChainId;
         string bridgeType;
     }
 
+    function constructBridgeOperations(
```
scott-silver, line 104: 🙌

src/builder/QuarkBuilder.sol
```diff
@@ -74,89 +72,67 @@ contract QuarkBuilder {
...
-                            sender: srcAccountBalances[j].account,
-                            // where it goes
-                            destinationChainId: transferIntent.chainId,
-                            recipient: transferIntent.sender,
-                            blockTimestamp: transferIntent.blockTimestamp
-                        }),
-                        payment
-                    );
+        // If action is paid for with tokens and the payment token is not the transfer token, attempt to bridge some over if not enough
+        // Note: The previous code block for bridging the transfer token already handles the case where payment token == transfer token
```
scott-silver, line 112: I'm confused by how this is structured. What are the two cases in which you would need to bridge?

kevincheng96: We have two bridging code blocks in the `transfer` function:

1. The first one (existing) bridges the transfer token.
2. The second one (new) bridges the payment token.

We only need to run the second code block if:

1. The action is paid for with tokens AND the payment token is not the transfer token. If the payment token is the transfer token, it would have been bridged by the first code block already.
2. There are not enough of the payment token on the destination chain (see `needsBridgedFunds` check)

src/builder/QuarkBuilder.sol
```diff
@@ -65,7 +63,7 @@ contract QuarkBuilder {
         TransferIntent memory transferIntent,
         Accounts.ChainAccounts[] memory chainAccountsList,
         PaymentInfo.Payment memory payment
-    ) external pure returns (BuilderResult memory) {
+    ) external view returns (BuilderResult memory) {
```
cwang25, line 66: Unintentional change from `pure` to `view`?

src/builder/QuarkBuilder.sol
```diff
@@ -16,16 +16,14 @@ contract QuarkBuilder {
...
 
     /* ===== Custom Errors ===== */
 
     error AssetPositionNotFound();
-    error FundsUnavailable(uint256 requiredAmount, uint256 actualAmount, uint256 missingAmount);
+    error FundsUnavailable(string assetSymbol, uint256 requiredAmount, uint256 actualAmount, uint256 missingAmount);
     error InsufficientFunds(uint256 requiredAmount, uint256 actualAmount);
     error InvalidInput();
     error MaxCostTooHigh();
-    error TooManyBridgeOperations();
```
cwang25, line 28: So we're dropping `TooManyBridgeOperations()` check?

src/builder/Actions.sol
```diff
@@ -84,25 +96,130 @@ library Actions {
...
+        IQuarkWallet.QuarkOperation[] memory quarkOperations =
+            new IQuarkWallet.QuarkOperation[](chainAccountsList.length);
+
+        // Note: Assumes that the asset uses the same # of decimals on each chain
+        uint256 balanceOnDstChain =
+            Accounts.getBalanceOnChain(bridgeInfo.assetSymbol, bridgeInfo.dstChainId, chainAccountsList);
+        uint256 amountLeftToBridge = bridgeInfo.amountNeededOnDst - balanceOnDstChain;
+        uint256 bridgeActionCount = 0;
+        // TODO: bridge routing logic (which bridge to prioritize, how many bridges?)
+        // Iterate chainAccountList and find up to 2 chains that can provide enough fund
```
cwang25, line 128: Is this comment still true that it runs upto 2 chains?

src/builder/QuarkBuilder.sol
kevincheng96: Good catch. I changed them to `view` because `console.log` doesn't work with `pure` functions. Will change these back.

src/builder/QuarkBuilder.sol
kevincheng96: Yeah, we discussed removing this earlier in the week. It's quite arbitrary and isn't really useful right now given there is only one supported bridge.

src/builder/Actions.sol
kevincheng96: Good catch, not anymore. Updated the comment

src/builder/QuarkBuilder.sol
```diff
@@ -16,16 +16,14 @@ contract QuarkBuilder {
     /* ===== Constants ===== */
 
     string constant VERSION = "1.0.0";
-    uint256 constant MAX_BRIDGE_ACTION = 1;
 
     /* ===== Custom Errors ===== */
 
     error AssetPositionNotFound();
-    error FundsUnavailable(uint256 requiredAmount, uint256 actualAmount, uint256 missingAmount);
+    error FundsUnavailable(string assetSymbol, uint256 requiredAmount, uint256 actualAmount, uint256 missingAmount);
```
scott-silver, line 23: I know this wasn't introduced in your PR, but `missingAmount` seems like an unnecessary argument.

It's always `requiredAmount - actualAmount`, right?

src/builder/Actions.sol
```diff
@@ -100,6 +111,7 @@ library Actions {
         uint256 price;
         address recipient;
         address token;
+        string assetSymbol;
```
scott-silver, line 114: nit: put in alphabetical order relative to the other fields (below `amount`)

src/builder/Actions.sol
```diff
@@ -152,6 +164,7 @@ library Actions {
...
```
scott-silver, line 167: same nit as above; put below `amount`

src/builder/QuarkBuilder.sol
scott-silver: I delete this function in #29, so one of our PRs is going to require a somewhat delicate rebase:

https://github.com/compound-finance/quark-scripts/pull/29/files#diff-9111113812791cbe6884c8fad436a6674788f2e44d1616c9bb124f30a33bd250L365

test/QuarkBuilder.t.sol
```diff
@@ -22,8 +22,11 @@ contract QuarkBuilderTest is Test {
     address constant ETH_USD_PRICE_FEED_8453 = 0x71041dddad3595F9CEd3DcCFBe3D1F4b0a16Bb70;
     address constant USDC_1 = 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48;
     address constant USDC_8453 = 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913;
-    // Random address for mock of USDC in random unsupported L2 chain
+    address constant USDT_1 = 0xdAC17F958D2ee523a2206206994597C13D831ec7;
+    address constant USDT_8453 = 0xfde4C96c8593536E31F229EA8f37b2ADa2699bb2;
+    // Random address for mock assets in random unsupported chain
     address constant USDC_7777 = 0x8D89c5CaA76592e30e0410B9e68C0f235c62B312;
+    address constant USDT_7777 = 0x8D89c5CaA76592e30e0410B9e68C0f235c62B312;
```
scott-silver, line 29: We should probably use a different random address than the one we're using for USDC, right?

Might cause issues, even if it's on our unsupported chain

scott-silver (approved):
💪

src/builder/QuarkBuilder.sol
kevincheng96: Yeah, I had the same thought. I can remove it

cwang25 (approved):

src/builder/Actions.sol
kevincheng96: Done

src/builder/Actions.sol
kevincheng96: Done

test/QuarkBuilder.t.sol
kevincheng96: Done

src/builder/QuarkBuilder.sol
kevincheng96: Done

PR #31, opened 2024-06-26
definitions for ActionContexts

scott-silver:
Definitions for our ActionContext structs.

Differences from previous specs on these contexts:
- `chainId` for all contexts
- "TransferContext" instead of "SendContext"
- "BridgeActionContext" is different (these changes already exist on `main`)
- added `token` to `SupplyActionContext` (so it's obvious which asset you're supplying)


hayesgm (commented):
lgtm

kevincheng96 (approved):

PR #34, opened 2024-06-28
Remove redundant `InsufficientFunds` check and properly account for amount needed on destination chain

kevincheng96:
We were not properly accounting for the payment token when checking if funds needed to be bridged. This only applies to the case where the payment token is also the action token. For example, if the user wants to transfer 10 USDC with a payment max cost of 1 USDC, the `needsBridgedFunds` needs to check that there is 11 USDC on the destination chain instead of just 10 USDC.

We also remove the `InsufficientFunds` check because it is redundant and seems like a subset of the `assertFundsAvailable` check.

### DIFF
--- a/test/builder/QuarkBuilderTransfer.t.sol
+++ b/test/builder/QuarkBuilderTransfer.t.sol
@@ -48,40 +48,53 @@ contract QuarkBuilderTransferTest is Test, QuarkBuilderTest {
 
     function testInsufficientFunds() public {
         QuarkBuilder builder = new QuarkBuilder();
-        vm.expectRevert(abi.encodeWithSelector(QuarkBuilder.InsufficientFunds.selector, 10e6, 0e6));
+        vm.expectRevert(abi.encodeWithSelector(QuarkBuilder.FundsUnavailable.selector, "USDC", 10e6, 0e6));
         builder.transfer(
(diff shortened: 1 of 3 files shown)


test/builder/QuarkBuilderTransfer.t.sol
```diff
@@ -48,40 +48,53 @@ contract QuarkBuilderTransferTest is Test, QuarkBuilderTest {
...
-            paymentUsdc_(maxCosts_(1, 1_000e6)) // but costs 1,000USDC
+            paymentUsdc_(maxCosts_(1, 1_000e6)) // but costs 1,000 USDC
         );
     }
 
-    function testFundsUnavailable() public {
+    function testFundsOnUnbridgeableChains() public {
         QuarkBuilder builder = new QuarkBuilder();
-        // FundsUnavailable(2e6, 0e6, 2e6): Requested 2e6, Available 0e6, Still missing 2e6
+        // FundsUnavailable(2e6, 0e6, 2e6): Requested 2e6, Available 0e6
```
scott-silver, line 73: Nit: I think that first argument in your comment should be the assetSymbol:

```
error FundsUnavailable(string assetSymbol, uint256 requiredAmount, uint256 actualAmount);
```

scott-silver (approved):

test/builder/QuarkBuilderTransfer.t.sol
kevincheng96: Done. Thanks

PR #35, opened 2024-07-02
Support Matcha swaps in QuarkBuilder

kevincheng96:
This PR adds support for Matcha swaps in the QuarkBuilder. I also changed the action context for swaps from two separate `Buy` and `Sell` action contexts to one single `Swap` action context.

### DIFF
--- a/src/builder/QuarkBuilder.sol
+++ b/src/builder/QuarkBuilder.sol
@@ -278,7 +278,7 @@ contract QuarkBuilder {
             }
         }
 
-        // Then, transferIntent `amount` of `assetSymbol` to `recipient`
+        // Then, transfer `amount` of `assetSymbol` to `recipient`
         (quarkOperations[actionIndex], actions[actionIndex]) = Actions.transferAsset(
             Actions.TransferAsset({
                 chainAccountsList: chainAccountsList,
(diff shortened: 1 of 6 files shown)


AungThuSoe24 (approved):

AungThuSoe24 (approved):

src/builder/QuarkBuilder.sol
```diff
@@ -332,6 +332,166 @@ contract QuarkBuilder {
...
+        actions = Actions.truncate(actions, actionIndex);
+        quarkOperations = Actions.truncate(quarkOperations, actionIndex);
+
+        // Validate generated actions for affordability
+        if (payment.isToken) {
+            assertSufficientPaymentTokenBalances(actions, chainAccountsList, swapIntent.chainId);
+        }
+
+        // Construct EIP712 digests
+        EIP712Helper.EIP712Data memory eip712Data;
```
hayesgm, line 469: We can share this code at some point, right? In fact, most of the end of this function could (technically, not sure about Solidity-wise) just be a common function call, right?

src/builder/QuarkBuilder.sol
```diff
@@ -332,6 +332,166 @@ contract QuarkBuilder {
...
+        if (quarkOperations.length == 1) {
+            eip712Data = EIP712Helper.EIP712Data({
+                digest: EIP712Helper.getDigestForQuarkOperation(
+                    quarkOperations[0], actions[0].quarkAccount, actions[0].chainId
+                    ),
+                domainSeparator: EIP712Helper.getDomainSeparator(actions[0].quarkAccount, actions[0].chainId),
+                hashStruct: EIP712Helper.getHashStructForQuarkOperation(quarkOperations[0])
+            });
+        } else if (quarkOperations.length > 1) {
```
hayesgm, line 478: What about the case, FWIW, where `quarkOperations.length == 0`?

hayesgm (approved):
dope dope dope

src/builder/QuarkBuilder.sol
kevincheng96: In that case, EIP712Data will be an empty struct.

src/builder/QuarkBuilder.sol
kevincheng96: Yeah, this logic is being used in multiple places now. We'll want to clean it up a bit in a separate PR.

PR #40, opened 2024-07-09
List helper lib

cwang25:
Small helper lib for list mainly for actions and quarkoperations list so that we don't have to "pre-guess" the elements every time we initialize those actions/operations batch.  Mainly a lib to help to work with something like dynamic arraylist type of use cases.
Also save some lines from tracking `index++` in various places and calling `truncat()` at the end of build.

### DIFF
--- a/src/builder/List.sol
+++ b/src/builder/List.sol
@@ -1,0 +1,91 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity 0.8.23;
+
+import {IQuarkWallet} from "quark-core/src/interfaces/IQuarkWallet.sol";
+import {Actions} from "./Actions.sol";
+
+library List {
+    error IndexOutOfBound();
+
+    struct ListStuct {
+        bytes[] bytesArray;
+        uint256 size;
+    }
+
+    function newList() internal pure returns (ListStuct memory) {
+        return ListStuct(new bytes[](0), 0);
+    }
+
+    function toArray(ListStuct memory list) internal pure returns (bytes[] memory) {
+        bytes[] memory result = new bytes[](list.size);
+        for (uint256 i = 0; i < list.size; ++i) {
+            result[i] = list.bytesArray[i];
+        }
+        return result;
+    }
+
+    function add(ListStuct memory list, bytes memory item) internal pure {
+        if ((list.size + 1) * 2 > list.bytesArray.length) {
+            bytes[] memory newBytesArray = new bytes[]((list.bytesArray.length + 1) * 2);
+            for (uint256 i = 0; i < list.size; ++i) {
+                newBytesArray[i] = list.bytesArray[i];
+            }
+            newBytesArray[list.size] = item;
+
+            list.bytesArray = newBytesArray;
+            list.size = list.size + 1;
+        } else {
+            list.bytesArray[list.size] = item;
+            list.size = list.size + 1;
+        }
+    }
+
+    function get(ListStuct memory list, uint256 index) internal pure returns (bytes memory) {
+        if (index >= list.size) revert IndexOutOfBound();
+        return list.bytesArray[index];
+    }
+
+    // Structs' APIs
+
+    // === QuarkOperations ===
+    function addQuarkOperation(ListStuct memory list, IQuarkWallet.QuarkOperation memory operation) internal pure {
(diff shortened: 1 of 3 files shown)


src/builder/List.sol
```diff
@@ -0,0 +1,91 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity 0.8.23;
+
+import {IQuarkWallet} from "quark-core/src/interfaces/IQuarkWallet.sol";
+import {Actions} from "./Actions.sol";
+
+library List {
+    error IndexOutOfBound();
+
+    struct ListStuct {
```
kevincheng96, line 10: Typo in `Struct`. But also, consider renaming this as `DynamicArray` since `List` is really generic 

```
    struct DynamicArray {
        bytes[] items;
        uint256 length;
    }
```

src/builder/List.sol
```diff
@@ -0,0 +1,91 @@
...
+            list.size = list.size + 1;
+        }
+    }
+
+    function get(ListStuct memory list, uint256 index) internal pure returns (bytes memory) {
+        if (index >= list.size) revert IndexOutOfBound();
+        return list.bytesArray[index];
+    }
+
+    // Structs' APIs
```
kevincheng96, line 48:
```suggestion
    // Struct APIs
```

src/builder/List.sol
```diff
@@ -0,0 +1,91 @@
...
+
+    // === QuarkOperations ===
```
kevincheng96, line 50: nit: extra line under this

src/builder/List.sol
```diff
@@ -0,0 +1,91 @@
...
+        returns (IQuarkWallet.QuarkOperation[] memory)
+    {
+        IQuarkWallet.QuarkOperation[] memory result = new IQuarkWallet.QuarkOperation[](list.size);
+        for (uint256 i = 0; i < list.size; ++i) {
+            result[i] = abi.decode(list.bytesArray[i], (IQuarkWallet.QuarkOperation));
+        }
+        return result;
+    }
+
+    // === Action ===
```
kevincheng96, line 75: nit: extra line under this

src/builder/List.sol
```diff
@@ -0,0 +1,91 @@
...
+
+    function toArray(ListStuct memory list) internal pure returns (bytes[] memory) {
+        bytes[] memory result = new bytes[](list.size);
+        for (uint256 i = 0; i < list.size; ++i) {
+            result[i] = list.bytesArray[i];
+        }
+        return result;
+    }
+
+    function add(ListStuct memory list, bytes memory item) internal pure {
```
kevincheng96, line 27: nit: addItem

src/builder/List.sol
```diff
@@ -0,0 +1,91 @@
...
+                newBytesArray[i] = list.bytesArray[i];
+            }
+            newBytesArray[list.size] = item;
+
+            list.bytesArray = newBytesArray;
+            list.size = list.size + 1;
+        } else {
+            list.bytesArray[list.size] = item;
+            list.size = list.size + 1;
+        }
```
kevincheng96, line 40: This can be simplified to something like:

```
  function addItem(DynamicArray memory array, bytes memory item) internal {
        if (array.items.length == array.length) {
            bytes[] memory newItems = new bytes[](array.length * 2 + 1);
            for (uint256 i = 0; i < array.length; ++i) {
                newItems[i] = array.items[i];
            }
            array.items = newItems;
        }
        array.items[array.length] = item;
        array.length++;
  }
```

src/builder/List.sol
kevincheng96: Note: `size` is a bit confusing because this is really the length of the array. The size is determined by `items.length`.

kevincheng96 (commented):
Nice, very clean 🧼

src/builder/List.sol
cwang25: Sure no problem 👍 
Oh yeah I think the reason I chose `size` initially was borrowing it from Java's Arraylist convention on size() haha. 
We can use `length` 😄

src/builder/List.sol
cwang25: thanks for the nice simplified version of the logics haha.

kevincheng96 (approved):
LGTM

PR #43, opened 2024-07-11
Merge Quark Operations on the same chain

kevincheng96:
If there are multiple Quark Operations for the same chain, we merge them into one operation that uses the Multicall script. We also delay wrapping of operations around paycall/quotecall until *after* the merging is done.


src/builder/QuarkBuilder.sol
```diff
@@ -79,6 +80,8 @@ contract QuarkBuilder {
             payment
         );
 
+        // TODO: set this properly
+        bool useQuotecall = false;
```
scott-silver, line 84: might be best to avoid unnecessary variables, to avoid `stack too deep` errors

src/builder/QuarkOperationHelper.sol
```diff
@@ -0,0 +1,150 @@
...
+import {PaymentInfo} from "./PaymentInfo.sol";
+import {QuotecallWrapper} from "./QuotecallWrapper.sol";
+
+// Helper library to merge QuarkOperations on the same chain
+library QuarkOperationHelper {
+    function mergeSameChainOperations(
+        IQuarkWallet.QuarkOperation[] memory quarkOperations,
+        Actions.Action[] memory actions
+    ) internal pure returns (IQuarkWallet.QuarkOperation[] memory, Actions.Action[] memory) {
+        // Note: Assumes quarkOperations and actions have the same length
```
scott-silver, line 20: should we revert if that's not the case?

src/builder/QuarkOperationHelper.sol
```diff
@@ -0,0 +1,150 @@
...
+        Actions.Action[][] memory groupedActions = new Actions.Action[][](uniqueChainCount);
+
+        // Initialize grouped arrays
+        for (uint256 i = 0; i < uniqueChainCount; ++i) {
+            groupedQuarkOperations[i] = new IQuarkWallet.QuarkOperation[](operationCounts[i]);
+            groupedActions[i] = new Actions.Action[](operationCounts[i]);
+        }
+
+        // Group operations by chain
+        uint256[] memory currentIndex = new uint256[](uniqueChainCount);
```
scott-silver, line 57: `currentIndices` since this is a list?

src/builder/QuarkOperationHelper.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    // Note: Assumes all the quark operations are for the same quark wallet.
+    function mergeOperations(IQuarkWallet.QuarkOperation[] memory quarkOperations, Actions.Action[] memory actions)
+        internal
+        pure
+        returns (IQuarkWallet.QuarkOperation memory, Actions.Action memory)
+    {
+        address[] memory callContracts = new address[](quarkOperations.length);
+        bytes[] memory callDatas = new bytes[](quarkOperations.length);
+        bytes[] memory scriptSources = new bytes[](quarkOperations.length + 1);
```
scott-silver, line 98: why an extra `scriptSource`?

src/builder/QuarkOperationHelper.sol
```diff
@@ -0,0 +1,150 @@
...
+            callContracts[i] = quarkOperations[i].scriptAddress;
+            callDatas[i] = quarkOperations[i].scriptCalldata;
+            // Note: For simplicity, we assume there is one script source per quark operation for now
+            scriptSources[i] = quarkOperations[i].scriptSources[0];
+        }
+        scriptSources[quarkOperations.length] = type(Multicall).creationCode;
+
+        bytes memory multicallCalldata = abi.encodeWithSelector(Multicall.run.selector, callContracts, callDatas);
+
+        // Construct QuarkOperation and Action
```
scott-silver, line 110: We don't actually construct an Action here, right?

src/builder/QuarkOperationHelper.sol
```diff
@@ -0,0 +1,150 @@
...
+        IQuarkWallet.QuarkOperation memory lastQuarkOperation = quarkOperations[quarkOperations.length - 1];
+        IQuarkWallet.QuarkOperation memory mergedQuarkOperation = IQuarkWallet.QuarkOperation({
+            nonce: lastQuarkOperation.nonce,
+            scriptAddress: CodeJarHelper.getCodeAddress(type(Multicall).creationCode),
+            scriptCalldata: multicallCalldata,
+            scriptSources: scriptSources,
+            expiry: lastQuarkOperation.expiry
+        });
+
+        return (mergedQuarkOperation, actions[actions.length - 1]);
```
scott-silver, line 122: do we discard all but the last action?

scott-silver (approved):
impressive 💪

src/builder/QuarkOperationHelper.sol
```diff
@@ -0,0 +1,150 @@
...
+// Helper library to merge QuarkOperations on the same chain
+library QuarkOperationHelper {
+    function mergeSameChainOperations(
+        IQuarkWallet.QuarkOperation[] memory quarkOperations,
+        Actions.Action[] memory actions
+    ) internal pure returns (IQuarkWallet.QuarkOperation[] memory, Actions.Action[] memory) {
+        // Note: Assumes quarkOperations and actions have the same length
+
+        // Arrays to keep track of unique chain IDs and their operations
+        uint256[] memory uniqueChainIds = new uint256[](quarkOperations.length);
```
cwang25, line 23: Part of me feel like if we want to have another helper lib something like `HashMap.sol` implementation for this. (something like List.sol)

src/builder/QuarkOperationHelper.sol
kevincheng96: Sure, added a check that reverts if the lengths are different

src/builder/QuarkOperationHelper.sol
kevincheng96: Yeah, we can consider that for a future PR

src/builder/QuarkOperationHelper.sol
kevincheng96: Done

src/builder/QuarkOperationHelper.sol
kevincheng96: Added a comment to explain this

src/builder/QuarkOperationHelper.sol
kevincheng96: Yeah we aren't constructing it, but we are picking an action (the last one) to return. Not sure what's a better way to convey it

src/builder/QuarkOperationHelper.sol
kevincheng96: Correct. As of now, that should only be the wrap actions, which are fine to discard. In the long run, we'll likely want to support multiple action contexts for a single quark operation. Right now, they are 1:1

cwang25 (approved):

PR #44, opened 2024-07-11
Start using semantic versioning in Quark Builder

kevincheng96:


scott-silver (approved):

cwang25 (approved):

hayesgm (approved):
Awesome you might want to consider starting in 0.0.1 until base stability. <1.0.0 you can change interfaces in all patches. >=1.0.0 breaking changes would require a major or minor patch.

kevincheng96:
> Awesome you might want to consider starting in 0.0.1 until base stability. <1.0.0 you can change interfaces in all patches. >=1.0.0 breaking changes would require a major or minor patch.

Make sense. I updated it to `0.0.1`.

PR #41, opened 2024-07-11
CometBorrow

scott-silver:

PR #45, opened 2024-07-12
CometRepay

scott-silver:


src/builder/Actions.sol
```diff
@@ -186,10 +187,11 @@ library Actions {
 
     struct RepayActionContext {
         uint256 amount;
+        string assetSymbol;
         uint256 chainId;
         uint256[] collateralAmounts;
-        uint256[] collateralTokenPrices;
-        address[] collateralTokens;
+        uint256[] collateralAssetPrices;
```
scott-silver, line 193: updating the action context names from `collateralToken_` to `collateralAsset_`

src/builder/QuarkBuilder.sol
```diff
@@ -52,6 +52,145 @@ contract QuarkBuilder {
...
+        // XXX confirm that the user is not withdrawing beyond their limits
+
+        assertFundsAvailable(
+            repayIntent.chainId, repayIntent.assetSymbol, repayIntent.amount, chainAccountsList, payment
+        );
+
+        uint256 actionIndex = 0;
+        Actions.Action[] memory actions = new Actions.Action[](chainAccountsList.length);
+        IQuarkWallet.QuarkOperation[] memory quarkOperations =
+            new IQuarkWallet.QuarkOperation[](chainAccountsList.length);
```
cwang25, line 84: Since the dynamic list is checked in, probably will be good to use dynamic list instead of tracking actionIndex in places.

src/builder/QuarkBuilder.sol
```diff
@@ -52,6 +52,145 @@ contract QuarkBuilder {
...
+
+        bool useQuotecall = false; // TODO: calculate an actual value for useQuoteCall
```
cwang25, line 86: Trying to think what would be a use case for quotecall in repay. 
I guess is when user has 10USDC,  and comet has -10USDC, 1ETH. 
So user trigger repayMax?

src/builder/QuarkBuilder.sol
```diff
@@ -1019,18 +1164,32 @@ contract QuarkBuilder {
             }
             paymentTokenCost += nonBridgeAction.paymentMaxCost;
 
-            if (Strings.stringEqIgnoreCase(nonBridgeAction.actionType, Actions.ACTION_TYPE_TRANSFER)) {
-                Actions.TransferActionContext memory transferActionContext =
-                    abi.decode(nonBridgeAction.actionContext, (Actions.TransferActionContext));
-                if (Strings.stringEqIgnoreCase(transferActionContext.assetSymbol, paymentTokenSymbol)) {
-                    paymentTokenCost += transferActionContext.amount;
+            if (Strings.stringEqIgnoreCase(nonBridgeAction.actionType, Actions.ACTION_TYPE_BORROW)) {
+                continue;
```
cwang25, line 1168: Why we don't have to worry about the cost for BORROW and WITHDRAW 🤔 ?

test/builder/lib/QuarkBuilderTest.sol
```diff
@@ -9,8 +9,11 @@ import {CodeJarHelper} from "src/builder/CodeJarHelper.sol";
...
 import {PaymentInfo} from "src/builder/PaymentInfo.sol";
 import {QuarkBuilder} from "src/builder/QuarkBuilder.sol";
+import {Strings} from "src/builder/Strings.sol";
 
 contract QuarkBuilderTest {
+    uint256 constant BLOCK_TIMESTAMP = 123_456_789;
+
     address constant COMET_1 = address(0xc3);
 
     address constant LINK_1 = address(0xfeed01);
```
cwang25, line 19: iirc XXX_# the # means the chain id. 
To prevent future confusion, probably it's better to use the actual addresses for those that are marked with the actual chain Id like 1 as mainnet?

src/builder/Actions.sol
scott-silver: will restore this to match #41

src/builder/QuarkBuilder.sol
scott-silver: will make that update in a follow-up PR; I want to get Borrow and Repay merged to unblock end-to-end testing

src/builder/QuarkBuilder.sol
scott-silver: because borrowing and/or withdrawing an asset increases your balance of that asset, whereas supplying that asset decreases your balance of that asset

kevincheng96 (approved):
LGTM

src/builder/Actions.sol
scott-silver: updated

PR #51, opened 2024-07-12
eip712DataForQuarkOperations helper function

scott-silver:
We're not currently testing the contents of the BuilderResult's eip712Data, but I haven't changed any of the logic here so the results should be identical as before


kevincheng96 (approved):
👍

PR #53, opened 2024-07-16
Migrate cometBorrow and cometRepay to use Dynamic list [p2]

cwang25:
- Prerequisite for adding auto-wrapper features to those two functions.

### DIFF
--- a/src/builder/QuarkBuilder.sol
+++ b/src/builder/QuarkBuilder.sol
@@ -78,10 +78,8 @@ contract QuarkBuilder {
             repayIntent.chainId, repayIntent.assetSymbol, repayIntent.amount, chainAccountsList, payment
         );
 
-        uint256 actionIndex = 0;
-        Actions.Action[] memory actions = new Actions.Action[](chainAccountsList.length);
-        IQuarkWallet.QuarkOperation[] memory quarkOperations =
-            new IQuarkWallet.QuarkOperation[](chainAccountsList.length);
+        List.DynamicArray memory actions = List.newList();
+        List.DynamicArray memory quarkOperations = List.newList();
 
         bool useQuotecall = false; // TODO: calculate an actual value for useQuoteCall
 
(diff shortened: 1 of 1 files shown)


src/builder/QuarkBuilder.sol
```diff
@@ -152,41 +151,43 @@ contract QuarkBuilder {
...
-                QuarkOperationHelper.wrapOperationsWithTokenPayment(quarkOperations, actions, payment, useQuotecall);
+            quarkOperationsArray = QuarkOperationHelper.wrapOperationsWithTokenPayment(
+                quarkOperationsArray, actionsArray, payment, useQuotecall
+            );
         }
 
         // Construct EIP712 digests
         EIP712Helper.EIP712Data memory eip712Data;
-        if (quarkOperations.length == 1) {
+        if (quarkOperationsArray.length == 1) {
```
scott-silver, line 171: This whole code block was deleted from `main` in #51. I'm guessing that if you merge p2 into p1 that you're going to have merge conflicts.

But if you rebase, you should be able to just delete any changes you've made here to lines 170-183.

src/builder/QuarkBuilder.sol
```diff
@@ -326,41 +325,43 @@ contract QuarkBuilder {
...
```
scott-silver, line 345: same comment as above; lines 343-359 no longer exist on main. You should be able to rebase and then remove those changes

src/builder/QuarkBuilder.sol
cwang25: Oh I see.  Thanks! I have rebased it so this change is now gone 👍

src/builder/QuarkBuilder.sol
cwang25: Oh I see.  Thanks! I have rebased it so this change is now gone 👍

kevincheng96 (approved):

PR #55, opened 2024-07-17
CometWithdraw max

scott-silver:
supporting "withdraw max" to `cometWithdraw` in QuarkBuilder.

supporting this feature requires that we add a representation of the caller's Comet positions, in order to know how much will be withdrawn, and therefore whether the caller has enough payment token to cover the cost of the transaction.

### DIFF
--- a/src/builder/Accounts.sol
+++ b/src/builder/Accounts.sol
@@ -12,7 +12,7 @@ library Accounts {
         uint256 chainId;
         QuarkState[] quarkStates;
         AssetPositions[] assetPositionsList;
-        CometPosition[] cometPositions;
+        CometPositions[] cometPositions;
     }
 
     // We map this to the Portfolio data structure that the client will already have.
@@ -38,7 +38,7 @@ library Accounts {
         uint256 balance;
     }
 
-    struct CometPosition {
+    struct CometPositions {
         address comet;
         CometBasePosition basePosition;
         CometCollateralPosition[] collaterals;
@@ -69,10 +69,10 @@ library Accounts {
         }
     }
 
-    function findCometPosition(uint256 chainId, address comet, ChainAccounts[] memory chainAccountsList)
+    function findCometPositions(uint256 chainId, address comet, ChainAccounts[] memory chainAccountsList)
         internal
         pure
-        returns (CometPosition memory found)
+        returns (CometPositions memory found)
     {
         ChainAccounts memory chainAccounts = findChainAccounts(chainId, chainAccountsList);
         for (uint256 i = 0; i < chainAccounts.cometPositions.length; ++i) {
--- a/src/builder/Accounts.sol
+++ b/src/builder/Accounts.sol
@@ -41,7 +41,7 @@ library Accounts {
     struct CometPositions {
         address comet;
         CometBasePosition basePosition;
-        CometCollateralPosition[] collaterals;
+        CometCollateralPosition[] collateralPositions;
     }
 
     struct CometBasePosition {
--- a/src/builder/Accounts.sol
+++ b/src/builder/Accounts.sol
@@ -12,6 +12,7 @@ library Accounts {
         uint256 chainId;
         QuarkState[] quarkStates;
         AssetPositions[] assetPositionsList;
+        CometPosition[] cometPositions;
     }
 
     // We map this to the Portfolio data structure that the client will already have.
@@ -37,6 +38,25 @@ library Accounts {
         uint256 balance;
     }
 
+    struct CometPosition {
+        address comet;
+        CometBasePosition basePosition;
+        CometCollateralPosition[] collaterals;
+    }
+
+    struct CometBasePosition {
+        address asset;
+        address[] accounts;
+        uint256[] borrowed;
+        uint256[] supplied;
+    }
+
+    struct CometCollateralPosition {
+        address asset;
+        address[] accounts;
+        uint256[] balances;
+    }
+
     function findChainAccounts(uint256 chainId, ChainAccounts[] memory chainAccountsList)
         internal
         pure
@@ -45,6 +65,19 @@ library Accounts {
         for (uint256 i = 0; i < chainAccountsList.length; ++i) {
             if (chainAccountsList[i].chainId == chainId) {
                 return found = chainAccountsList[i];
+            }
+        }
+    }
+
+    function findCometPosition(uint256 chainId, address comet, ChainAccounts[] memory chainAccountsList)
+        internal
+        pure
+        returns (CometPosition memory found)
+    {
+        ChainAccounts memory chainAccounts = findChainAccounts(chainId, chainAccountsList);
+        for (uint256 i = 0; i < chainAccounts.cometPositions.length; ++i) {
+            if (chainAccounts.cometPositions[i].comet == comet) {
+                return found = chainAccounts.cometPositions[i];
             }
         }
     }
--- a/src/builder/Actions.sol
+++ b/src/builder/Actions.sol
@@ -708,11 +708,11 @@ library Actions {
         return (quarkOperation, action);
     }
 
-    function cometWithdrawAsset(CometWithdraw memory cometWithdraw, PaymentInfo.Payment memory payment)
-        internal
-        pure
-        returns (IQuarkWallet.QuarkOperation memory, Action memory)
-    {
+    function cometWithdrawAsset(
+        CometWithdraw memory cometWithdraw,
+        PaymentInfo.Payment memory payment,
+        bool useQuotecall
+    ) internal pure returns (IQuarkWallet.QuarkOperation memory, Action memory) {
         bytes[] memory scriptSources = new bytes[](1);
         scriptSources[0] = type(CometWithdrawActions).creationCode;
 
@@ -756,7 +756,7 @@ library Actions {
             quarkAccount: cometWithdraw.withdrawer,
             actionType: ACTION_TYPE_WITHDRAW,
             actionContext: abi.encode(cometWithdrawActionContext),
-            paymentMethod: payment.isToken ? PAYMENT_METHOD_PAYCALL : PAYMENT_METHOD_OFFCHAIN,
+            paymentMethod: paymentMethodForPayment(payment, useQuotecall),
             // Null address for OFFCHAIN payment.
             paymentToken: payment.isToken
                 ? PaymentInfo.knownToken(payment.currency, cometWithdraw.chainId).token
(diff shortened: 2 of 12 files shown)


src/builder/Accounts.sol
```diff
@@ -12,6 +12,7 @@ library Accounts {
         uint256 chainId;
         QuarkState[] quarkStates;
         AssetPositions[] assetPositionsList;
+        CometPosition[] cometPositions;
```
scott-silver, line 15: This structure is a subset of the `cometPositions` data that is already being generated by Portfolios. It's not necessarily the optimal way to structure this data, but I think we're aiming for doing whatever is simplest for the existing client.

src/builder/QuarkBuilder.sol
```diff
@@ -501,17 +501,15 @@ contract QuarkBuilder {
...
         CometWithdrawIntent memory cometWithdrawIntent,
         Accounts.ChainAccounts[] memory chainAccountsList,
         PaymentInfo.Payment memory payment
     ) external pure returns (BuilderResult memory) {
         // XXX confirm that you actually have the amount to withdraw
 
-        // TODO: set this properly
-        bool useQuotecall = false;
+        bool isMaxWithdraw = cometWithdrawIntent.amount == type(uint256).max;
+        bool useQuotecall = payment.isToken && isMaxWithdraw;
```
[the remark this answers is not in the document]
scott-silver: using Paycall for all max withdraws is easy.

I guess the more important part of this PR is establishing a structure for passing in information on a user's Comet positions (which will still be necessary for `cometWithdraw` since we need to know how much of the payment token the user will have after the operation has been executed).

src/builder/QuarkBuilder.sol
```diff
@@ -574,8 +573,24 @@ contract QuarkBuilder {
...
+                    //
+                    // assumes that a maxWithdraw is for the base asset, since
+                    // you cannot maxWithdraw Collateral
+                    Accounts.CometPosition memory cometPosition = Accounts.findCometPosition(
+                        cometWithdrawIntent.chainId, cometWithdrawIntent.comet, chainAccountsList
+                    );
+
+                    for (uint256 i = 0; i < cometPosition.basePosition.accounts.length; ++i) {
+                        if (cometPosition.basePosition.accounts[i] == cometWithdrawIntent.withdrawer) {
+                            supplementalPaymentTokenBalance += cometPosition.basePosition.supplied[i];
```
scott-silver, line 588: here we're using `basePosition.supplied` to get the best guess of how much of the asset the user will end up with ultimately.

i'm assuming that `basePosition.supplied` is best way piece of information we have about the current withdrawable amount, but I don't know if that number reflects how much of the base asset has been supplied or the amount supplied + interest accrued

```
"cometPositions" => [
    %{
      "comet" => "0x2943ac1216979ad8db76d9147f64e61adc126e96",
      "borrowAPR" => 9945224338320000,
      "supplyAPR" => 4793472000,
      "borrowRewardsAPR" => 0,
      "supplyRewardsAPR" => 1699636062375893083,
      "base" => %{
        "asset" => "0x2d5ee574e710219a521449679a4a7f2b43f046ad",
        "decimals" => 18,
        "usdPrice" => 315288151686,
        "totalBorrow" => 969017000499,
        "totalSupply" => 3405644257260925148,
        "totalBalance" => 3405643280745890843,
        "accounts" => [
          "0xf6F148cFb92dD3baF399B0eCD1a8561b7f16487D",
          "0xE276Bc378A527A8792B353cdCA5b5E53263DfB9e"
        ],
        "borrowed" => [0, 0],
        "supplied" => [0, 0],
        "usdBorrowCapacity" => [0, 0],
        "baseBorrowCapacity" => [0, 0],
        "usdLiquidationPoint" => [0, 0],
        "baseLiquidationPoint" => [0, 0]
      },
      "collaterals" => [
        %{
          "asset" => "0xb9fa8f5ec3da13b508f462243ad0555b46e028df",
          "decimals" => 18,
          "usdPrice" => 305829507135,
          "basePrice" => 97000000,
          "totalSupply" => 2000134115105271783,
          "accounts" => [
            "0xf6F148cFb92dD3baF399B0eCD1a8561b7f16487D",
            "0xE276Bc378A527A8792B353cdCA5b5E53263DfB9e"
          ],
          "balances" => [0, 0]
        },
      ]
    }
 ]
```

src/builder/QuarkBuilder.sol
```diff
@@ -574,8 +573,24 @@ contract QuarkBuilder {
         if (payment.isToken) {
             uint256 supplementalPaymentTokenBalance = 0;
             if (Strings.stringEqIgnoreCase(payment.currency, cometWithdrawIntent.assetSymbol)) {
-                // XXX in the withdrawMax case, use the Comet balance
-                supplementalPaymentTokenBalance += cometWithdrawIntent.amount;
+                if (isMaxWithdraw) {
+                    // when doing a maxWithdraw of the payment token, add the account's supplied balance
+                    // as supplemental payment token balance
+                    //
+                    // assumes that a maxWithdraw is for the base asset, since
```
scott-silver, line 580: I did not include a check that the withdrawn token is the base asset. using an amount of `uint256.max` when withdrawing a collateral asset will revert.

this is also a check we can do in the as-yet-implemented check against your CometPositions, which we is unblocked now that we're passing in CometPosition info.

src/builder/Accounts.sol
kevincheng96, line 15: `CometPositions` with an `s`?

src/builder/Accounts.sol
```diff
@@ -37,6 +38,25 @@ library Accounts {
         uint256 balance;
     }
 
+    struct CometPosition {
+        address comet;
+        CometBasePosition basePosition;
+        CometCollateralPosition[] collaterals;
```
kevincheng96, line 44: nit: `collateralPositions`

src/builder/Actions.sol
```diff
@@ -837,6 +830,20 @@ library Actions {
         return (quarkOperation, action);
     }
 
+    function paymentMethodForPayment(PaymentInfo.Payment memory payment, bool useQuotecall)
```
kevincheng96, line 833: Great to pull this out into a separate helper. This can live in the `PaymentInfo` library.

kevincheng96: Also, there's some other functions in this file that this can be used in.

src/builder/QuarkBuilder.sol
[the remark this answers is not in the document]
kevincheng96: Agree that this can always be Paycall, so we can set this as `false` or remove the local variable altogether. I'd keep it for now just for readability and remove it if we run into stack too deep errors.

src/builder/QuarkBuilder.sol
kevincheng96: Yeah it probably makes more sense to check that in the check against the user's CometPositions. So it's fair to assume `isMaxWithdraw` is for the base asset, as the earlier check should revert if not.

src/builder/QuarkBuilder.sol
kevincheng96: I assume it includes interest accrued, as there's no way to get the original supplied amount in the current block.

test/builder/lib/QuarkBuilderTest.sol
```diff
@@ -260,13 +281,61 @@ contract QuarkBuilderTest {
...
+            });
+        }
+
+        return cometPositions;
+    }
+
+    function baseAssetForComet(uint256 chainId, address comet) internal pure returns (address) {
+        if (comet == COMET_1_USDC || comet == COMET_8453_USDC) {
+            return usdc_(chainId);
+        } else {
```
kevincheng96, line 334: Missing a case for WETH markets

test/builder/QuarkBuilderCometWithdraw.t.sol
```diff
@@ -373,4 +376,129 @@ contract QuarkBuilderCometWithdrawTest is Test, QuarkBuilderTest {
...
+        assertEq(result.actions[0].quarkAccount, address(0xa11ce), "0xa11ce sends the funds");
+        assertEq(result.actions[0].actionType, "WITHDRAW", "action type is 'WITHDRAW'");
+        assertEq(result.actions[0].paymentMethod, "QUOTE_CALL", "payment method is 'QUOTE_CALL'");
+        assertEq(result.actions[0].paymentToken, USDC_1, "payment token is USDC");
+        assertEq(result.actions[0].paymentMaxCost, 0.1e6, "payment max is set to 0.1e6");
+        assertEq(
+            result.actions[0].actionContext,
+            abi.encode(
+                Actions.WithdrawActionContext({
+                    amount: type(uint256).max, // ?? should this be a different amount?
```
kevincheng96, line 486: Good question. Let's check with :kelp: on this. Though there really isn't a way for QuarkBuilder to get the exact amount because of interest accrual.

kevincheng96 (approved):
Great work 💯

src/builder/Accounts.sol
scott-silver: I went with `collaterals` since that's what the existing backend calls that field (I failed to be consistent, however, and renamed `base` to `basePosition` 🤦 ).

Does it matter whether the field names are the same as the backend?

```
"cometPositions" => [
    %{
      "comet" => "0x2943ac1216979ad8db76d9147f64e61adc126e96",
      "base" => %{
        "asset" => "0x2d5ee574e710219a521449679a4a7f2b43f046ad",
        "accounts" => [
          "0xf6F148cFb92dD3baF399B0eCD1a8561b7f16487D",
          "0xE276Bc378A527A8792B353cdCA5b5E53263DfB9e"
        ],
        "borrowed" => [0, 0],
        "supplied" => [0, 0],]
      },
      "collaterals" => [
        %{
          "asset" => "0xb9fa8f5ec3da13b508f462243ad0555b46e028df",
          "accounts" => [
            "0xf6F148cFb92dD3baF399B0eCD1a8561b7f16487D",
            "0xE276Bc378A527A8792B353cdCA5b5E53263DfB9e"
          ],
          "balances" => [0, 0]
        },
      ]
    }
 ]
```

src/builder/Accounts.sol
scott-silver: It feels slightly weird for the type itself to be plural. Like an array of addresses would be `address[]` not `addresses[]`.

But I guess the argument for `CometPositions` is that each instance of the type itself (potentially) contains multiple positions, like multiple accounts with multiple `base` balances?

src/builder/QuarkBuilder.sol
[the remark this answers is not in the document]
scott-silver: updated to use Paycall

src/builder/Accounts.sol
scott-silver: updated to `CometPositions`

src/builder/Actions.sol
scott-silver: moved into `PaymentInfo` and updated the other functions in `Actions` to use it

test/builder/lib/QuarkBuilderTest.sol
scott-silver: added

test/builder/QuarkBuilderCometWithdraw.t.sol
scott-silver: @mykelp can you confirm that it is acceptable to return the `WithdrawActionContext` with `uint256.max` as the value for a max withdraw?

mykelp: yeah, that makes sense. We can handle it appropriately in the view

kevincheng96 (approved):

cwang25 (approved):

src/builder/Accounts.sol
scott-silver: updated to `collateralPositions`

PR #56, opened 2024-07-17
Hashmap helper lib to simplify the merge operation function

cwang25:
Hashmap helper lib to abstract away the need of using multiple array to track on unique ids. To give us somewhat an ability to use basic mapping functionality while in pure environment.

### DIFF
--- a/src/builder/HashMap.sol
+++ b/src/builder/HashMap.sol
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: BSD-3-Clause
 pragma solidity 0.8.23;
 
 import {List} from "./List.sol";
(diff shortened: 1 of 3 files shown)


src/builder/HashMap.sol
```diff
@@ -0,0 +1,123 @@
...
+    struct Map {
+        List.DynamicArray entries;
+    }
+
+    function newMap() internal pure returns (Map memory) {
+        return Map(List.newList());
+    }
+
+    function get(Map memory map, bytes memory key) internal pure returns (bytes memory) {
+        for (uint256 i = 0; i < map.entries.length; i++) {
```
kevincheng96, line 24: nit: ++1 just for consistency throughout this file

kevincheng96 (approved):
Sweet, looks great!

PR #57, opened 2024-07-18
CometRepay max

scott-silver:
Implementing max repay for `cometRepay`

### DIFF
--- a/src/builder/QuarkBuilder.sol
+++ b/src/builder/QuarkBuilder.sol
@@ -26,7 +26,6 @@ contract QuarkBuilder {
     error AssetPositionNotFound();
     error FundsUnavailable(string assetSymbol, uint256 requiredAmount, uint256 actualAmount);
     error InvalidActionChain();
-    error InvalidActionType();
     error InvalidInput();
     error MaxCostTooHigh();
     error MissingWrapperCounterpart();
@@ -74,22 +73,28 @@ contract QuarkBuilder {
 
         // XXX confirm that the user is not withdrawing beyond their limits
 
-        assertFundsAvailable(
-            repayIntent.chainId, repayIntent.assetSymbol, repayIntent.amount, chainAccountsList, payment
-        );
+        bool isMaxRepay = repayIntent.amount == type(uint256).max;
+        bool useQuotecall = isMaxRepay;
+
+        uint256 repayAmount;
+        if (isMaxRepay) {
+            repayAmount += Accounts.totalBorrowForAccount(
+                chainAccountsList, repayIntent.chainId, repayIntent.comet, repayIntent.repayer
+            );
+            // XXX apply buffer to repayAmount
+            repayAmount = repayAmount + 0;
+        } else {
+            repayAmount = repayIntent.amount;
+        }
+
+        assertFundsAvailable(repayIntent.chainId, repayIntent.assetSymbol, repayAmount, chainAccountsList, payment);
 
         List.DynamicArray memory actions = List.newList();
         List.DynamicArray memory quarkOperations = List.newList();
 
-        bool useQuotecall = false; // TODO: calculate an actual value for useQuoteCall
-
-        if (
-            needsBridgedFunds(
-                repayIntent.assetSymbol, repayIntent.amount, repayIntent.chainId, chainAccountsList, payment
-            )
-        ) {
+        if (needsBridgedFunds(repayIntent.assetSymbol, repayAmount, repayIntent.chainId, chainAccountsList, payment)) {
             // Note: Assumes that the asset uses the same # of decimals on each chain
-            uint256 amountNeededOnDst = repayIntent.amount;
+            uint256 amountNeededOnDst = repayAmount;
             // If action is paid for with tokens and the payment token is the
             // repay token, we need to add the max cost to the
             // amountNeededOnDst for target chain
@@ -123,7 +128,7 @@ contract QuarkBuilder {
             chainAccountsList,
             payment,
             repayIntent.assetSymbol,
-            repayIntent.amount,
+            repayAmount,
             repayIntent.chainId,
             repayIntent.repayer,
             repayIntent.blockTimestamp,
@@ -143,7 +148,8 @@ contract QuarkBuilder {
                 blockTimestamp: repayIntent.blockTimestamp,
                 repayer: repayIntent.repayer
             }),
-            payment
+            payment,
+            useQuotecall
         );
 
         List.addAction(actions, repayActions);
@@ -157,15 +163,20 @@ contract QuarkBuilder {
         if (payment.isToken) {
             // if you are withdrawing the payment token, you can pay with the
             // withdrawn funds
-            uint256 supplementalPaymentTokenBalance = 0;
+            uint256 paymentTokenReceived = 0;
             for (uint256 i = 0; i < repayIntent.collateralAssetSymbols.length; ++i) {
                 if (Strings.stringEqIgnoreCase(payment.currency, repayIntent.collateralAssetSymbols[i])) {
-                    supplementalPaymentTokenBalance += repayIntent.collateralAmounts[i];
+                    paymentTokenReceived += repayIntent.collateralAmounts[i];
                 }
             }
 
+            uint256 paymentTokenSpent = 0;
+            if (Strings.stringEqIgnoreCase(repayIntent.assetSymbol, payment.currency)) {
+                paymentTokenSpent += repayAmount;
+            }
+
             assertSufficientPaymentTokenBalances(
-                actionsArray, chainAccountsList, repayIntent.chainId, supplementalPaymentTokenBalance
+                actionsArray, chainAccountsList, repayIntent.chainId, paymentTokenReceived, paymentTokenSpent
             );
         }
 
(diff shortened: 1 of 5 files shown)


src/builder/QuarkBuilder.sol
```diff
@@ -74,22 +73,28 @@ contract QuarkBuilder {
 
         // XXX confirm that the user is not withdrawing beyond their limits
 
-        assertFundsAvailable(
-            repayIntent.chainId, repayIntent.assetSymbol, repayIntent.amount, chainAccountsList, payment
-        );
+        bool isMaxRepay = repayIntent.amount == type(uint256).max;
+        bool useQuotecall = isMaxRepay;
```
scott-silver, line 77: this should always be true, right? whenever we're doing a max repay we should use Quotecall

src/builder/QuarkBuilder.sol
```diff
@@ -74,22 +73,28 @@ contract QuarkBuilder {
...
+
+        uint256 repayAmount;
+        if (isMaxRepay) {
+            repayAmount += Accounts.totalBorrowForAccount(
+                chainAccountsList, repayIntent.chainId, repayIntent.comet, repayIntent.repayer
+            );
+            // XXX apply buffer to repayAmount
```
scott-silver, line 84: Not sure what our buffer logic should be.

Maybe we add either 1% of the `repayAmount` or some `MIN_BUFFER` value, whichever is larger?

src/builder/QuarkBuilder.sol
```diff
@@ -123,7 +128,7 @@ contract QuarkBuilder {
             chainAccountsList,
             payment,
             repayIntent.assetSymbol,
-            repayIntent.amount,
+            repayAmount,
```
scott-silver, line 131: 🪄 can you confirm: the value passed to `checkAndInsertWrapOrUnwrapAction` should be our best estimate of the actual repayAmount and not simply `type(uint256).max`, correct?

src/builder/QuarkBuilder.sol
```diff
@@ -1100,25 +1139,35 @@ contract QuarkBuilder {
...
     function assertSufficientPaymentTokenBalances(
         Actions.Action[] memory actions,
         Accounts.ChainAccounts[] memory chainAccountsList,
         uint256 targetChainId
     ) internal pure {
-        return assertSufficientPaymentTokenBalances(actions, chainAccountsList, targetChainId, 0);
+        return assertSufficientPaymentTokenBalances(actions, chainAccountsList, targetChainId, 0, 0);
+    }
+
+    function assertSufficientPaymentTokenBalances(
```
scott-silver, line 1156: The changes to `assertSufficientPaymentTokenBalances` are the biggest change in this PR.

The structure of this function assumes that you can determine how much of the payment token is spent in an action based on its `ActionContext`. Frustratingly, that's not the case with a max Repay, because the ActionContext's `amount` is `type(uint256).max`.

So I've refactored this function to not read values from the various ActionContexts but instead to take in `paymentTokenReceived` and `paymentTokenSpent` arguments, so the caller can pass in their calculation of how much of the payment token was gained or lost in the actions.

The one part of the calculation that I've not recreated in this refactoring is the amount of the payment token that is spent and/or received by any Wrapping/Unwrapping actions that are added. Technically, I think we could do it, but it seems unnecessary as I don't think we're going to be doing any wrapping/unwrapping of USDC.

scott-silver: I'm open to other suggestions if this refactor seems too dramatic.

src/builder/QuarkBuilder.sol
cwang25: Yes. 
The logics within `checkAndInsertWrapOrUnwrapAction` is essentially trying to wrap/unwrap counter part of meet the input amount. So it will break if it try to wrap/unwrap int256.max of ETH or WETH. Unless we modify max handling in `checkAndInsertWrapOrUnwrapAction` to support it. 

iirc `checkAndInsertWrapOrUnwrapAction` should be put in the last statement insert "right" before the actual action is generated. By then, we should have a number. (such as transferMax, at this stage, it should has user total blanace amount pass into transferAction)

src/builder/QuarkBuilder.sol
cwang25: I think it's more like...
```
Case 1: 
User owed 100 USDC to Comet. User hold 50USDC. 
User use maxRepay, it used out all user's 50USDC to pay the debt and make it owe 50USDC at the end. 

In this case Quotecall is better. (Similar to transferMax to Comet.)

Case 2: 
User owe 100USDC to comet. User hold 150USDC.
User use RepayMax, it uses user holding of 100USDC and payoff all the debt at Comet and leaving 50USDC in wallet.

In this case use Paycall is better, as there is no need to drain some dust from the wallet to make user feel "empty" wallet.  (User still has 50USDC in the wallet anyway.)
```

To decide when to use quotecall, I think mostly goes down to whether the use case we want user to feel better when they see their "emptied" wallet.

scott-silver: Wouldn't Case 1 revert? If you called `maxRepay` for a 100 USDC borrow position but you only had 50 USDC?

cwang25: If only case 2, then probably just paycall should be good. Cuz Case 2 we don't have to make user feel "emptied" 🤔 

Was thinking about builder to smart switch between apply all user owed position v.s. apply user maxBalance if user try to do maxRepay when not having all the fund.

src/builder/QuarkBuilder.sol
cwang25: That sounds good to me 👍 
Or we can just do % increase as well,  since interest accrue is purely based on %.

src/builder/QuarkBuilder.sol
cwang25: hmm yeah, and it feels like part of the `assertSufficientPaymentTokenBalance` logics gets removed outside of the function and scattered into various places in above code blocks.  Would prefer to make that function to covers most cases, so people can simply put that check function at the end assuming it can help reviewing the final answer and tell you if you can afford it or not. 

If all action needs to calculate paymentToken on their own, then probably `wrapOrUnwrap` needs to be refactor in case of wrap/unwrap away the paymentToken cases.

Wondering if it's hard to replicate the math done in line166 in this function to calculate here instead 🤔

scott-silver: We can't do the math here with the existing arguments, because there is no piece of data in `chainListAccounts` or `actions` that captures what the actual repayAmount is going to be. The `amount` argument in the `ActionContext` is just `uint256.max`.

The user's comet positions are available as part of `chainListAccounts`, but you don't know which account to look for the max borrow amount for.

src/builder/QuarkBuilder.sol
kevincheng96: Weirdly enough, I think this is a case where we always use `Paycall`. In transfer max, all the funds leave the user's wallet, so `Quotecall` is used to not leave any dust. In repay max, the user is not transferring all the funds in their wallet. They are transferring just enough to pay back their entire borrow. So `Quotecall` here doesn't actually achieve anything.

src/builder/QuarkBuilder.sol
kevincheng96: Yeah % makes sense. Something small like 0.1% is probably already plenty for most situations, but this really depends on the current borrow rate, which might be too complex to build in for now. Maybe leave that as a TODO.

Just to highlight how large 0.1% already is, let's use 20% borrow APR as an example. If we convert this to a daily rate, we get 20/365=0.0548%. So 0.1% pays for 2 days of a borrow that has 20% borrow APR. That seems plenty if we expect to repay to happen in an hour or so.

src/builder/QuarkBuilder.sol
```diff
@@ -1100,25 +1139,35 @@ contract QuarkBuilder {
...
     }
 
     function assertSufficientPaymentTokenBalances(
         Actions.Action[] memory actions,
         Accounts.ChainAccounts[] memory chainAccountsList,
         uint256 targetChainId,
-        uint256 supplementalPaymentTokenBalance
+        uint256 paymentTokenReceived,
+        uint256 paymentTokenSpent
     ) internal pure {
```
kevincheng96, line 1171: Nice, this interface is pretty clean and understandable

kevincheng96 (approved):
Looks great. Main two things to address:

- `Paycall` instead of `Quotecall`
- % buffer to add to repay amount

src/builder/QuarkBuilder.sol
cwang25: I see.

src/builder/QuarkBuilder.sol
```diff
@@ -1160,52 +1209,11 @@ contract QuarkBuilder {
...
-                }
-            } else if (
-                Strings.stringEqIgnoreCase(nonBridgeAction.actionType, Actions.ACTION_TYPE_UNWRAP)
-                    || Strings.stringEqIgnoreCase(nonBridgeAction.actionType, Actions.ACTION_TYPE_WRAP)
-            ) {
-                Actions.WrapOrUnwrapActionContext memory wrapOrUnwrapActionContext =
-                    abi.decode(nonBridgeAction.actionContext, (Actions.WrapOrUnwrapActionContext));
-                if (Strings.stringEqIgnoreCase(wrapOrUnwrapActionContext.fromAssetSymbol, paymentTokenSymbol)) {
-                    paymentTokenCost += wrapOrUnwrapActionContext.amount;
-                }
```
cwang25, line 1198: I think once this logics are removed, so I assume all this logics will be handled separately in above steps. 
So probably need to add this to each wrapOrUnwrap step as well. 
(When wrap/unwrap away from payment token (pay with ETH, wrapping ETH away to WETH), the cost need to adjust accordingly)

scott-silver:
Closing in favor of #58

PR #58, opened 2024-07-23
CometRepay max, v2

scott-silver:
This is a reworked version of #57 ; I've done it as a separate PR so we can compare the two approaches.

The downside of the approach in #57, as 🪄 correctly pointed out, is that the cost of wrapping/unwrapping assets would have to be incorporated into the `paymentTokenSpent` argument passed to `assertSufficientPaymentTokenBalances`. This would be logic that would have to be reimplemented by every function that used `assertSufficientPaymentTokenBalances`.

So instead, this PR maintains `assertSufficientPaymentTokenBalances` largely as it is, but adds an additional `account` argument. With `account`, `assertSufficientPaymentTokenBalances` can calculate the borrow amount for a max repay.
 
(This PR also incorporates the last pieces of feedback from #57: always using `Paycall` and adding a 0.1% buffer to the repay amount for max repays)

### DIFF
--- a/src/builder/QuarkBuilder.sol
+++ b/src/builder/QuarkBuilder.sol
@@ -1197,11 +1197,10 @@ contract QuarkBuilder {
                 Actions.RepayActionContext memory cometRepayActionContext =
                     abi.decode(nonBridgeAction.actionContext, (Actions.RepayActionContext));
                 if (Strings.stringEqIgnoreCase(cometRepayActionContext.assetSymbol, paymentTokenSymbol)) {
-                    // weird variable placement to avoid stack-too-deep error
-                    uint256 repayAmount = cometRepayMaxAmount(
-                        chainAccountsList, cometRepayActionContext.chainId, cometRepayActionContext.comet, account
-                    );
                     if (cometRepayActionContext.amount == type(uint256).max) {
+                        uint256 repayAmount = cometRepayMaxAmount(
+                            chainAccountsList, cometRepayActionContext.chainId, cometRepayActionContext.comet, account
+                        );
                         paymentTokenCost += repayAmount;
                     } else {
                         paymentTokenCost += cometRepayActionContext.amount;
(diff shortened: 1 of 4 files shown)


src/builder/QuarkBuilder.sol
```diff
@@ -74,22 +85,25 @@ contract QuarkBuilder {
 
         // XXX confirm that the user is not withdrawing beyond their limits
 
-        assertFundsAvailable(
-            repayIntent.chainId, repayIntent.assetSymbol, repayIntent.amount, chainAccountsList, payment
-        );
+        bool isMaxRepay = repayIntent.amount == type(uint256).max;
+        bool useQuotecall = false;
```
kevincheng96, line 89: nit: can add the same comment here: `// never use Quotecall`

src/builder/QuarkBuilder.sol
```diff
@@ -1167,7 +1197,14 @@ contract QuarkBuilder {
                 Actions.RepayActionContext memory cometRepayActionContext =
                     abi.decode(nonBridgeAction.actionContext, (Actions.RepayActionContext));
                 if (Strings.stringEqIgnoreCase(cometRepayActionContext.assetSymbol, paymentTokenSymbol)) {
-                    paymentTokenCost += cometRepayActionContext.amount;
```
kevincheng96, line 1170: Unrelated to this line, but can you add a TODO on L1227 to test logic of wrap/unwrap affecting the `paymentTokenCost`? Basically, 🪄 made a good catch and we should have a test that breaks if the code is removed accidentally.

kevincheng96 (approved):
Cool, makes sense. IMO, this is more preferable than having to handle the wrap/unwrap  payment cost logic in each function.

kevincheng96:
Also, you'll want to up the builder version to "0.0.4". I increased it to "0.0.3" in a recent PR.

src/builder/QuarkBuilder.sol
scott-silver: added

PR #62, opened 2024-08-07
Rename expectedBuyAmount to buyAmount

kevincheng96:
Since the swap functions can support both exact input and exact output swaps, the buy amount is not always an expected amount. It can be the actual buy amount in an exact output swap.


hayesgm (commented):
looks good-- curious how many places we need to change code to match this 😢

kevincheng96:
> looks good-- curious how many places we need to change code to match this 😢

It should only require a client-level change. We don't touch the action context here.

hayesgm:
Yeah, though in both Elixir and Swift now. Well, let's see it out, a small change should be an easier one to understand.

hayesgm (approved):

PR #63, opened 2024-08-09
Recurring Swap script

kevincheng96:
`RecurringSwap.sol` is a Quark script that allows a Quark wallet to swap for an asset at a regular cadence. For example, one could submit an order to "Buy 1000 USDC worth of ETH every week". 

The current implementation is built with the following features:

- User can set an amount of asset A to buy using an asset B
- User can set a fixed interval for buying the asset
- User can cancel the recurring buy
- The recurring swap has oracle price checks to protect against MEV


src/RecurringPurchase.sol
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /// @notice The set of parameters for performing a swap
+    struct SwapParams {
+        address uniswapRouter;
+        address recipient;
+        address tokenFrom;
+        uint256 amount;
+        uint256 amountInMaximum; // Optional, for "exact out" swaps
+        uint256 amountOutMinimum; // Optional, for "exact in" swaps
```
kevincheng96, line 42: Actually, these won't work for swaps scheduled in the future because they are static. I think we may have to resort to checking the slippage against an oracle (Chainlink or TWAP) as 🌽 suggested.

src/RecurringPurchase.sol
```diff
@@ -0,0 +1,270 @@
...
+        allowReplay();
+
+        if (config.slippageParams.priceFeeds.length == 0) {
+            revert InvalidInput();
+        }
+        if (config.slippageParams.priceFeeds.length != config.slippageParams.shouldReverses.length) {
+            revert InvalidInput();
+        }
+
+        bytes32 hashedConfig = _hashConfig(config);
```
cwang25, line 98: Looks like here stores a timestamp with hashedConfig as a key to handle time interval, and the config has below unique fields.
```
config.swapParams.uniswapRouter,
config.swapParams.recipient,
config.swapParams.tokenIn,
config.swapParams.tokenOut,
config.swapParams.amount,
config.swapParams.isExactOut,
config.swapParams.deadline,
config.swapParams.path
```

If user added `buy 1ETH every day` then 1 day later, user created another recurring buy request and it's also `buy 1 ETH every day`

Would those configs conflict?  (those two seems will result in exact same set of parameters. Deadline are probably is uint256.max all the time?)  In this case, I think expect behavior will be like two same recurring buy running in parallel.


Probably, one way we could solve this is we "maintain" the `deadline` as always to be unique to the time of the first submission. Or add a additional field that's unique on the time of first submission probably. 🤔

src/RecurringPurchase.sol
```diff
@@ -0,0 +1,270 @@
...
+    /// @notice Parameters for controlling slippage in a swap operation
+    struct SlippageParams {
+        /// @dev Maximum acceptable slippage, expressed as a percentage where 100% = 1e18
+        uint256 maxSlippage;
+        /// @dev Price feed addresses for determining market exchange rates between token pairs
+        /// Example: For SUSHI -> SNX swap, use [SUSHI/ETH feed, SNX/ETH feed]
+        address[] priceFeeds;
+        /// @dev Flags indicating whether each corresponding price feed should be inverted
+        /// Example: For USDC -> ETH swap, use [true] with [ETH/USD feed] to get ETH per USDC
+        bool[] shouldReverses;
```
cwang25, line 76: nit: should we do a struct of array instead?

cwang25 (commented):
looks good to me.  I'm just a bit wary about the potential config conflict 🤔

src/RecurringPurchase.sol
kevincheng96: That new recurring purchase request will be a different Quark Operation with its own nonce right? If so, there will be no storage conflict because the storage in `StateManager` is isolated and nonce-scoped.

src/RecurringPurchase.sol
kevincheng96: Yeah, an array of structs could work as well. It's kind of a trade-off because we'll introduce another layer of nested structs, which is annoying to deal with in solidity. So I'm not completely sold on either approach, but can be swayed if someone has a stronger opinion on it.

src/RecurringPurchase.sol
cwang25: oh I see.
So in that case we probably don't need to hash a key to read but even some constant string value key would work too?

cwang25 (approved):
LGTM 

Discussed offline with ⛑️, only thing that need to be caution is the precision loss in math division chain in pricefeed.  But I believe most of the time will be fine.

src/RecurringPurchase.sol
fluffywaffles: Prior experience lands me on the side of "struct-of-arrays". So I would leave this as-is. :+1:

hayesgm (approved):
We should obv. get through audit, and it's going to change via new Replayable code, burt generally makes sense as is right now.

src/RecurringPurchase.sol
fluffywaffles, line 76: Nit: `shouldReverses` strikes me as odd wording; I think `shouldInvert` would be fine. I guess I also prefer "invert" over "reverse," but :shrug:

src/RecurringPurchase.sol
```diff
@@ -0,0 +1,270 @@
...
+    struct SwapParams {
+        address uniswapRouter;
+        address recipient;
+        address tokenIn;
+        address tokenOut;
+        /// @dev The amount for tokenIn if exact in; the amount for tokenOut if exact out
+        uint256 amount;
+        /// @dev False for exact in; true for exact out
+        bool isExactOut;
+        uint256 deadline;
```
fluffywaffles, line 63: unit? I'm assuming seconds

src/RecurringPurchase.sol
```diff
@@ -0,0 +1,270 @@
...
+
+    /**
+     * @dev Note: This script uses the following storage layout in the QuarkStateManager:
+     *         mapping(bytes32 hashedPurchaseConfig => uint256 nextPurchaseTime)
+     *             where hashedPurchaseConfig = keccak256(PurchaseConfig)
+     */
+
+    /// @notice Parameters for a recurring purchase order
+    struct PurchaseConfig {
+        uint256 interval;
```
fluffywaffles, line 48: unit? also presumably seconds

might consider being explicit there: `intervalSeconds` is not too cumbersome.

src/RecurringPurchase.sol
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /**
+     * @notice Scales an amount from one decimal precision to another decision precision
+     * @param amount The amount to be scaled
+     * @param fromDecimals The number of decimals in the source precision
+     * @param toDecimals The number of decimals in the target precision
+     * @return The scaled amount adjusted to the target precision
+     */
+    function _scaleDecimals(uint256 amount, uint256 fromDecimals, uint256 toDecimals) internal pure returns (uint256) {
```
fluffywaffles, line 237: Suuper tiny nit: kinda feel like this is more clear if called `rescale`.

test/RecurringPurchase.t.sol
```diff
@@ -0,0 +1,786 @@
...
+    address constant USDC_USD_PRICE_FEED = 0x8fFfFfd4AfB6115b954Bd326cbe7B4BA576818f6;
+
+    constructor() {
+        // Fork setup
+        vm.createSelectFork(
+            string.concat(
+                "https://node-provider.compound.finance/ethereum-mainnet/", vm.envString("NODE_PROVIDER_BYPASS_KEY")
+            ),
+            18429607 // 2023-10-25 13:24:00 PST
+        );
```
fluffywaffles, line 60: Note that we no longer control `node-provider.compound.finance` and eventually cannot rely on having access to bypass keys for tests.

fluffywaffles (approved):
tests look pretty thorough, too :+1: :+1:

src/RecurringPurchase.sol
kevincheng96: Agreed, `shouldReverses` is odd sounding. Changed it to `shouldInvert`.

src/RecurringPurchase.sol
kevincheng96: Yeah it's in seconds. I usually would attach units to these, but in Solidity I think it's always assumed to be in seconds unless otherwise stated. For example, `expiry` is used widely without specifying seconds. I'll leave a `@dev` comment to make sure there's no confusion.

src/RecurringPurchase.sol
kevincheng96: Similar comment here.

src/RecurringPurchase.sol
```diff
@@ -0,0 +1,270 @@
...
+    function cancel() external {
+        // Not explicitly clearing the nonce just cancels the replayable txn
+    }
+
+    /**
+     * @notice Execute a swap given a configuration for a recurring purchase
+     * @param config The configuration for a recurring purchase order
+     */
+    function purchase(PurchaseConfig calldata config) public {
+        allowReplay();
```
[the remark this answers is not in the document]
kevincheng96: Hmm, I do remember that about one of the `allowX` methods, but also don't quite remember the exact details. Though I also have a vague recollection it had to do with callbacks, rather than replays. I could look through old PRs to try to dig it up, but I'm glad we're getting rid of `allowReplay()` in v2 so it's no longer relevant in this context. Though, it would still be good to figure out what that exact edge-case is so we could document it in `QuarkScripts.sol`.

kevincheng96 (commented):
Apart from addressing the PR comments, I added a few more changes in the latest commit:

* Rename from `RecurringPurchase` to `RecurringSwap` 
* Introduce a `startTime` in `SwapConfig`. This allows for recurring swaps that are scheduled to start in the future
* Multiply intermediate values by `PRECISION_FACTOR = 1e18` to reduce precision loss when multiplying/dividing price feeds
* Rename custom error `PurchaseConditionNotMet` to a more specific `SwapWindowNotOpen(uint256 nextSwapTime, uint256 currentTime)`

src/RecurringPurchase.sol
fluffywaffles: comment is totallly sufficient 👍

PR #64, opened 2024-08-15
add numbers to too much slippage and too expensive

coburncoburn:
This patch adds more information to some difficult to debug reverts.


src/Paycall.sol
```diff
@@ -17,7 +17,7 @@ contract Paycall {
 
     error BadPrice();
     error InvalidCallContext();
-    error TransactionTooExpensive();
+    error TransactionTooExpensive(uint256 expectedMaxCost, uint256 actualMaxCost);
```
kevincheng96, line 20: Similar change should be made to `Quotecall` as well

src/lib/DeFiScriptErrors.sol
```diff
@@ -10,5 +10,5 @@ library DeFiScriptErrors {
     error InvalidInput();
     error TransferFailed(bytes data);
     error ApproveAndSwapFailed(bytes data);
-    error TooMuchSlippage();
+    error TooMuchSlippage(uint256 actualBuyAmount, uint256 expectedBuyAmount);
```
kevincheng96, line 13: nit: we should flip the order to be consistent with `TransactionTooExpensive(uint256 expectedMaxCost, uint256 actualMaxCost)`, which has expected first, then actual

kevincheng96:
Looks like there is a failing unit test

fluffywaffles (commented):
Agree with ⛑️  that we should stick to `(expected, actual)`.

Overall LGTM but will hold ✅ for ⛑️ 's comments.

src/lib/DeFiScriptErrors.sol
coburncoburn: doh, yes

coburncoburn:
> Looks like there is a failing unit test

weird this test passes locally and in fact is explicitly the test that I changed...

.github/workflows/test.yml
kevincheng96, line 1: Should we revert these changes?

kevincheng96 (approved):
Look good. Weird that snapshot is failing though. Maybe it needs to be re-gen'ed again?

.github/workflows/test.yml
coburncoburn: yeah the forge fmt changes are too much for me to stomach today

kevincheng96 (approved):

PR #65, opened 2024-08-22
Support recurring swaps in QuarkBuilder

kevincheng96:
This implements the QuarkBuilder logic to construct Quark Operations for recurring swaps. We introduce a few new libraries that hardcode some addresses: `PriceFeeds.sol` and `UniswapRouter.sol`.


src/builder/QuarkBuilder.sol
```diff
@@ -1092,7 +1213,35 @@ contract QuarkBuilder {
...
+        string memory assetSymbol,
+        uint256 amount,
+        uint256 chainId,
+        Accounts.ChainAccounts[] memory chainAccountsList,
+        PaymentInfo.Payment memory payment
+    ) internal pure returns (bool) {
+        uint256 balanceOnChain = getBalanceOnChain(assetSymbol, chainId, chainAccountsList, payment);
+        uint256 amountNeededOnChain = getAmountNeededOnChain(assetSymbol, amount, chainId, payment);
+
+        return balanceOnChain < amountNeededOnChain;
```
kevincheng96, line 1244: These diffs are just pulling out the logic in `needsBridgedFunds` into two separate helper functions.

banky (approved):
Looks good ✅

src/builder/PriceFeeds.sol
```diff
@@ -0,0 +1,284 @@
...
+            allFeeds[currentIndex] = baseSepoliaFeeds[i];
+            currentIndex++;
+        }
+
+        return allFeeds;
+    }
+
+    // Mainnet
+    function knownPriceFeeds_1() internal pure returns (PriceFeed[] memory) {
+        PriceFeed[] memory priceFeeds = new PriceFeed[](4);
```
banky, line 51: Do we plan to add every price feed for all supported tokens here?

kevincheng96: Yeah, we'll have to hardcode all the price feeds for tokens we care about. 🧇 is working on a crawler that will output a Solidity config file that will replace this file.

PR #66, opened 2024-08-22
Morpho integration (p1) - Defi script

cwang25:
- Basic interfaces to interact with Morpho Blue (morpho), MetaMorpho (Morpho vault), Morpho reward contracts


src/DeFiScripts.sol
kevincheng96, line 1: Since this file is getting pretty large, I think it's better to create a separate file specifically for Morpho actions (`MorphoScript.sol`). Eventually, we may want to pull the Comet actions out as well.

src/interfaces/IMetaMorpho.sol
kevincheng96, line 1: Are these Morpho interfaces taken directly from somewhere? If so, would be good to leave a `@dev` comment to link to the source.

src/DeFiScripts.sol
```diff
@@ -327,3 +330,111 @@ contract ApproveAndSwap {
         IERC20(sellToken).forceApprove(to, 0);
     }
 }
+
+// === Morpho actiosn ===
+contract MorphoVaultActions {
```
kevincheng96, line 335: These contracts and functions are missing natspecs.

src/interfaces/IMorpho.sol
```diff
@@ -0,0 +1,41 @@
...
+    address oracle;
+    address irm;
+    uint256 lltv;
+}
+
+struct Position {
+    uint256 supplyShares;
+    uint128 borrowShares;
+    uint128 collateral;
+}
```
kevincheng96, line 41: Should these be defined inside the interface?

src/DeFiScripts.sol
```diff
@@ -327,3 +330,111 @@ contract ApproveAndSwap {
...
+            IERC20(marketParams.loanToken).forceApprove(morpho, assets);
+            (assetsRepaid, sharesRepaid) = IMorpho(morpho).repay(marketParams, assets, shares, onBehalf, data);
+        } else {
+            IERC20(marketParams.loanToken).forceApprove(morpho, type(uint256).max);
+            (assetsRepaid, sharesRepaid) = IMorpho(morpho).repay(marketParams, assets, shares, onBehalf, data);
+            IERC20(marketParams.loanToken).forceApprove(morpho, 0);
+        }
+    }
+
+    function supplyCollateral(
```
kevincheng96, line 392: Similar to our Comet functions, we probably want a way to `Supply collateral and borrow` in a Quark script.  Same with `Repay and withdraw collateral`.

src/DeFiScripts.sol
```diff
@@ -327,3 +330,111 @@ contract ApproveAndSwap {
...
+
+    function repay(
+        address morpho,
+        MarketParams memory marketParams,
+        uint256 assets,
+        uint256 shares,
+        address onBehalf,
+        bytes calldata data
+    ) external returns (uint256 assetsRepaid, uint256 sharesRepaid) {
+        if (assets > 0) {
```
kevincheng96, line 382: Maybe we should assert that only one of these (`assets` and `shares`) can be non-zero.

kevincheng96 (commented):
Nice, looking great so far

src/DeFiScripts.sol
cwang25: I totally agreed. Will split into MorphoScript.sol 👍 
That DefiScripts has became more like Misc.sol

src/interfaces/IMetaMorpho.sol
cwang25: Will add @dev comment,  it's a mixture of IMorpho and the ERC4262 interfaces that it's inheriting from.

src/interfaces/IMorpho.sol
cwang25: Yeah, those are directly from the Morpho that we could use to decode their return. 
Probably the same format as IComet.sol.

src/DeFiScripts.sol
cwang25: oh yeah haha.
There was a moment I wanted to put this PR with another PR as one. 
These two actions got mistakenly made into the second PR 😓 instead of this one.
I think I will just move that changes to here, and remove that from the second PR then.

src/DeFiScripts.sol
cwang25: Sure, although morpho contracts doesn't check that, I think for morpho contracts if it sees shares# it will replace whatever passed in `assets`

src/defi_integrations/MorphoScripts.sol
```diff
@@ -0,0 +1,261 @@
...
+    function repayAndWithdrawCollateral(
+        address morpho,
+        MarketParams memory marketParams,
+        uint256 repayAmount,
+        uint256 repayShares,
+        uint256 withdrawAmount,
+        address onBehalf,
+        address receiver
+    ) external {
+        if (repayAmount > 0 || repayShares > 0) {
```
kevincheng96, line 176: This outer if-statement seems unnecessary

src/defi_integrations/MorphoScripts.sol
kevincheng96, line 1: Let's use named parameters for multi-arity functions.

kevincheng96, line 1: I don't think we necessarily need a new subdirectory called `defi_integrations` right now. Or if we do add a subdirectory, I don't think `defi_integrations` is the best way to modularize our directories, since I would say most of the current quark scripts are related to defi.

kevincheng96 (approved):

kevincheng96 (changes requested):
Oops, didn't mean to give the approval just now

src/defi_integrations/MorphoScripts.sol
cwang25: Sure can remove that folder. 
My plan was as time goes we will have multiple fields with different defi names, and was thinking to put those in this folder such as Comet, Morpho, Aave, Sommelier...etc

src/defi_integrations/MorphoScripts.sol
```diff
@@ -0,0 +1,261 @@
...
+     * @param vault The address of the MetaMorpho vault
+     * @param shares The amount of shares to redeem
+     * @return assets The amount of assets redeemed
+     */
+    function redeem(address vault, uint256 shares) external returns (uint256) {
+        return IMetaMorpho(vault).redeem(shares, address(this), address(this));
+    }
+}
+
+contract MorphoBlueActions {
```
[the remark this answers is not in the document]
cwang25: Yeah ... it's a bit annoying that they keep changing names haha. (Makersao > sky 😂)

 I ended up following on their actual contract names that's deployed on etherscan which is still Morpho Blue. 

I think now it's like 
MorphoBlue/Morpho 
MetaMorpho/Morpho Vault 
😵‍💫

src/defi_integrations/MorphoScripts.sol
cwang25: Was mainly for saving actions on builder that client can simply just withdraw collateral by passing zero repay amount so the code could skip the repay clause without having to call separate script.

src/defi_integrations/MorphoScripts.sol
```diff
@@ -0,0 +1,261 @@
...
+    ) external returns (uint256 assetsRepaid, uint256 sharesRepaid) {
+        if (assets > 0) {
+            IERC20(marketParams.loanToken).forceApprove(morpho, assets);
+            (assetsRepaid, sharesRepaid) = IMorpho(morpho).repay(marketParams, assets, shares, onBehalf, data);
+        } else if (shares > 0) {
+            IERC20(marketParams.loanToken).forceApprove(morpho, type(uint256).max);
+            (assetsRepaid, sharesRepaid) = IMorpho(morpho).repay(marketParams, assets, shares, onBehalf, data);
+            IERC20(marketParams.loanToken).forceApprove(morpho, 0);
+        } else {
+            revert InvalidInput();
```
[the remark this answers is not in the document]
cwang25: ok. Actually there was a moment I want to deprecated DefiScriptErrors haha. 
Since now its in different files, not really sure if we should group all those error in one random shared solidity file 🤔 
(But I guess that's for ease of parsing from client/backend side?)

src/defi_integrations/MorphoScripts.sol
cwang25: Which functions? 
You mean the function calls to Morpho interface?

src/DeFiScripts.sol
```diff
@@ -9,6 +9,9 @@ import {QuarkScript} from "quark-core/src/QuarkScript.sol";
 
 import {IComet} from "./interfaces/IComet.sol";
 import {ICometRewards} from "./interfaces/ICometRewards.sol";
+import {IMorpho, MarketParams} from "./interfaces/IMorpho.sol";
+import {IMetaMorpho} from "./interfaces/IMetaMorpho.sol";
+import {IMorphoUniversalRewardsDistributor} from "./interfaces/IMorphoUniversalRewardsDistributor.sol";
```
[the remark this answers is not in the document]
cwang25: removed

src/defi_integrations/MorphoScripts.sol
cwang25: changed all the one with named parameters.  Left some positional parameters that's is straightforward.

src/defi_integrations/MorphoScripts.sol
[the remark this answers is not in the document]
cwang25: Changed to Morpho. 
i guess repo name is just repo name, and the solidity file is named `morpho` lol

src/defi_integrations/MorphoScripts.sol
cwang25: The latest logics has been adjusted to: 
```
if (repayAmount > 0 || repayShares > 0) {
            if (repayAmount == type(uint256).max) {
                // Repay max
                IERC20(marketParams.loanToken).forceApprove(morpho, type(uint256).max);
                IMorpho(morpho).repay({});
                IERC20(marketParams.loanToken).forceApprove(morpho, 0);
            } else if (repayAmount > 0) {
                IERC20(marketParams.loanToken).forceApprove(morpho, repayAmount);
                IMorpho(morpho).repay({ });
            } else {
                IERC20(marketParams.loanToken).forceApprove(morpho, type(uint256).max);
                IMorpho(morpho).repay({});
                IERC20(marketParams.loanToken).forceApprove(morpho, 0);
            }
        }
```

So outer if became more like if user gives both 0, it's likely user just want to withdraw collateral,  so outer if skip the whole repay actions clause.

test/MorphoVaultActions.t.sol
```diff
@@ -67,27 +67,6 @@ contract MorphoVaultActionsTest is Test {
         assertApproxEqAbs(IERC4626(morphoVault).balanceOf(address(wallet)), 9713.4779e18, 0.01e18);
```
kevincheng96, line 67: Would be good to also assert that `IERC4626.convertToAssets(shares) == 10_000e6`.

test/MorphoVaultActions.t.sol
kevincheng96, line 1: I can't comment on specific lines, but wanted to say:

nit: there should be a extra line between "act" and "assert" blocks in tests. 

e.g. there should be an extra line above

```
        vm.resumeGasMetering();
        wallet.executeQuarkOperation(op, v, r, s);
```

Same for the other test file

kevincheng96 (approved):
LGTM!

test/MorphoVaultActions.t.sol
cwang25: Sounds good.  But due to nature of rounding down on share conversion, in this case there will be 1 wei diff.  
So will need to use Approx here too.

test/MorphoVaultActions.t.sol
cwang25: Sure add extra lines above and below of the execute lines to separate out the before and after assert blocks.

src/MorphoScripts.sol
```diff
@@ -0,0 +1,193 @@
...
+     * @param marketParams The market parameters of the individual morpho blue market
+     * @param repayAmount The amount of assets to repay, pass in `type(uint256).max` to repay max
+     * @param repayShares The amount of shares to repay
+     * @param withdrawAmount The amount of assets to withdraw as collateral
+     */
+    function repayAndWithdrawCollateral(
+        address morpho,
+        MarketParams memory marketParams,
+        uint256 repayAmount,
+        uint256 repayShares,
```
fluffywaffles, line 64: we abstract over shares in every other action method, but not this one -- is there a particular reason why we expose the shares concept for `repayAndWithdrawCollateral`?

It seems from  a straightforward read of the code that there's one path that cares about `repayShares`, where we do not specify an exact asset amount to repay. But from my understanding of morpho, the only likely case where we would specify shares is repay max -- and we encode that using `type(uint256).max` amounts.

As it stands, we are also not doing anything to catch it if a caller specifies both amounts. This is ambiguous, and probably an error? Ideally it shouldn't even be possible to specify the same amount two ways, since it creates these invalid states.

Looking at the builder code in #67, I see that we always encode the calldata with `0` as repay shares. (In fact 0 is hardcoded.) So it would appear that this codepath is unused in practice. I see we have one test-case for repay-with-shares, but this is actually just the "repay max" test scenario, which is already covered by the `type(uint256).max` case.

Should we not just abstract over shares entirely, and remove `repayShares` from this method?

test/MorphoActions.t.sol
```diff
@@ -0,0 +1,194 @@
...
+        assertEq(IERC20(wstETH).balanceOf(address(wallet)), 0);
+        assertApproxEqAbs(
+            IMorpho(morpho).position(marketId(marketParams), address(wallet)).borrowShares, 9.533e14, 0.1e14
+        );
+
+        vm.resumeGasMetering();
+        wallet.executeQuarkOperation(op, v, r, s);
+
+        assertEq(IERC20(USDC).balanceOf(address(wallet)), 200e6);
+        assertEq(IERC20(wstETH).balanceOf(address(wallet)), 5e18);
```
fluffywaffles, line 83: From my read of this test case, we supplied `10wstETH` and borrowed `1000USDC`, then repaid `800USDC` and withdrew `5wstETH`. So should we not assert that the morpho balance for this address is still `5wstETH` of collateral and a remaining borrow of `200USDC`?

For the collateral balance, we test those in what looks like a straightforward way on line 146 in the supply-nd-borrow test-case. So that should be easy.

For the borrow balance, I think here, as above, we should actually compute our expectation, perhaps using `SharesMathLib`.

test/MorphoActions.t.sol
```diff
@@ -0,0 +1,194 @@
...
+        // Repay 800 USDC and withdraw 5 wstETH collateral
+        QuarkWallet.QuarkOperation memory op = new QuarkOperationHelper().newBasicOpWithCalldata(
+            wallet,
+            MorphoActionsScripts,
+            abi.encodeWithSelector(
+                MorphoActions.repayAndWithdrawCollateral.selector, morpho, marketParams, 800e6, 0, 5e18
+            ),
+            ScriptType.ScriptSource
+        );
+        (uint8 v, bytes32 r, bytes32 s) = new SignatureHelper().signOp(alicePrivateKey, wallet, op);
```
fluffywaffles, line 72: it's confusing that this op is created here but not executed until line 80. Can we move this down so that the asertions just below immediately follow the code that actually creates the conditions?

test/MorphoActions.t.sol
```diff
@@ -0,0 +1,194 @@
...
+        assertEq(IERC20(USDC).balanceOf(address(wallet)), 1000e6);
+        assertEq(IERC20(wstETH).balanceOf(address(wallet)), 0);
+        assertApproxEqAbs(
+            IMorpho(morpho).position(marketId(marketParams), address(wallet)).borrowShares, 9.533e14, 0.1e14
```
fluffywaffles, line 76: It bothers me a little bit that we're not actually computing our expectation here. Where does `9.533e14` come from? We could be using [`SharesMathLib`](https://github.com/morpho-org/morpho-blue/blob/main/src/libraries/SharesMathLib.sol) to explicitly calculate these expectations, and I would feel like I understood the test better and trusted the results more.

test/MorphoActions.t.sol
```diff
@@ -0,0 +1,194 @@
...
+        );
+
+        vm.resumeGasMetering();
+        wallet.executeQuarkOperation(op, v, r, s);
+
+        assertEq(IERC20(USDC).balanceOf(address(wallet)), 200e6);
+        assertEq(IERC20(wstETH).balanceOf(address(wallet)), 5e18);
+    }
+
+    function testRepayAndWithdrawColalteralInShares() public {
```
fluffywaffles, line 86: Mentioned above, but I don't really think we should expose the "in-shares" version of repay-and-withdraw, and therefore we don't need this test case.

test/MorphoActions.t.sol
```diff
@@ -0,0 +1,194 @@
...
+        // Supply 10 wstETH as collateral and borrow 1000 USDC
+        QuarkWallet.QuarkOperation memory op = new QuarkOperationHelper().newBasicOpWithCalldata(
+            wallet,
+            MorphoActionsScripts,
+            abi.encodeWithSelector(
+                MorphoActions.supplyCollateralAndBorrow.selector, morpho, marketParams, 10e18, 1000e6
+            ),
+            ScriptType.ScriptSource
+        );
+        (uint8 v, bytes32 r, bytes32 s) = new SignatureHelper().signOp(alicePrivateKey, wallet, op);
```
fluffywaffles, line 138: again, I find it confusing that we construct the op and it is immediately followed by assertions that are _pre_-conditions. I would prefer if we move this code down 3 lines and have the assertions of the "before" state come before the op.

test/MorphoRewardsActions.t.sol
```diff
@@ -0,0 +1,136 @@
...
+
+import {DeFiScriptErrors} from "src/lib/DeFiScriptErrors.sol";
+
+import "src/DeFiScripts.sol";
+import "src/MorphoScripts.sol";
+
+/**
+ * Tests for Morpho Rewards Claim
+ */
+contract MorphoRewardsActionsTest is Test {
```
fluffywaffles, line 30: Nb., but Morpho rewards is so confusing. This test is confusing! The actual work happens off-chain and the claims process is an awkward shuffling dance coordinating so many contracts and addresses. Blah.

In effect this test is asserting that "if Morpho's rewards API gives us these magic numbers, we get out some other magic numbers." Yet I don't see how we can do better! :cry:

test/MorphoVaultActions.t.sol
```diff
@@ -0,0 +1,98 @@
...
+        assertEq(IERC4626(morphoVault).balanceOf(address(wallet)), 0);
+
+        vm.resumeGasMetering();
+        wallet.executeQuarkOperation(op, v, r, s);
+
+        assertEq(IERC20(USDC).balanceOf(address(wallet)), 0);
+        assertApproxEqAbs(IERC4626(morphoVault).balanceOf(address(wallet)), 9713.4779e18, 0.01e18);
+        assertApproxEqAbs(
+            IERC4626(morphoVault).convertToAssets(IERC4626(morphoVault).balanceOf(address(wallet))), 10_000e6, 0.01e6
+        );
```
fluffywaffles, line 72: So maybe we don't need to use morpho's `SharesMathLib`, but we should do this! In the other tests.

Although I still prefer we don't hard-code expected shares counts in these assertions (the one above, in this case, on line 69), because to someone reading the test that number is meaningless. It would be a lot better to compute expected share counts, just like in this case we compute that the shares, converted back into an asset balance, are very nearly the same as the amount supplied.

test/MorphoVaultActions.t.sol
```diff
@@ -0,0 +1,98 @@
...
+        (uint8 v, bytes32 r, bytes32 s) = new SignatureHelper().signOp(alicePrivateKey, wallet, op);
+        assertEq(IERC20(USDC).balanceOf(address(wallet)), 0e6);
+        assertEq(IERC4626(morphoVault).balanceOf(address(wallet)), 10_000e18);
+
+        vm.resumeGasMetering();
+        wallet.executeQuarkOperation(op, v, r, s);
+
+        assertEq(IERC20(USDC).balanceOf(address(wallet)), 10_000e6);
+        assertApproxEqAbs(IERC4626(morphoVault).balanceOf(address(wallet)), 286.5e18, 1e18);
```
fluffywaffles, line 96: Again, as with other assertions regarding shares, I'm not sure how this assertion is supposed to convey information to me if I see that this test is failing.

Can we at least add a string annotation as the final call parameter? `"Expected exact shares dust in vault"` or something, perhaps.

fluffywaffles (commented):
Left a few notes, but concerning tests in particular:

1. We should try to avoid hardcoding shares counts where we have the option to do a little bit of shares math and turn the test into something legible to a reader. I bring this up in a few places. Even if we keep an assertion that is denominated in shares, we should add an assertion denominated in the equivalent asset amounts right next to it. Otherwise, as a reader of the code, I have no idea how to interpret those abstract shares numbers, and I won't understand the context if ever the assertion fails.

2. When/if we do end up using shares numbers directly in assertions, let's add a note-string to the assertion so that if and when it fails the person reading the failure has some idea of what exactly went wrong. This should be possible by passing a final argument to any `assert*` method which is just a string of the assertion note.

src/MorphoScripts.sol
cwang25: And yes we don't use repayShares now, as we only want to use it when we want to repay max. 
Since scripts handles repay max we don't need to use this field anymore. 

But leave repayShares here as MorphoScripts is just scripts to interact with Morpho, so since they open that options up, so I didn't completely trim off the whole interfaces they had. 

But agree, we can just remove it.


For both amounts, I just passed it to Morpho contract and let them handle it. 
As far as how Morpho handles it  is (amounts over shares / shares over amount) and won't throw error.

test/MorphoRewardsActions.t.sol
cwang25: Yes and people can claim for other wallets too. 
This opens up as we might can provide a service to batch claim for users.

test/MorphoVaultActions.t.sol
cwang25: Yeah it would be perfect. 
Unfortunately only MetaMorpho is IERC4626 compliance 😄

test/MorphoActions.t.sol
cwang25: I sort of understand the reason behind this one that do all the generations including `op` beforehand. Then we group all the assertions at the end with one of the execution to split `before` and `after`.

For that I think let's do it in a separate PR as designated as `QoL PR: test refactor`
Because most of our current of the existing tests are doing in this order too,  so it would be better to just move everyone to the same ordering would be better imo. To be consistent across whole suites.

test/MorphoActions.t.sol
cwang25: Sounds good will add those checks.

test/MorphoActions.t.sol
fluffywaffles: I do not intend to kick off a large refactor across all our tests.

It's fine to drop this if we just don't want to make the change.

src/MorphoScripts.sol
fluffywaffles: > But agree, we can just remove it.

I think that's best, unless we're actually going to use shares-based accounting in some way.

> For both amounts, I just passed it to Morpho contract and let them handle it. As far as how Morpho handles it is (amounts over shares / shares over amount) and won't throw error.

Yeah, but we have an opportunity abstracting over Morpho to avoid situations like this, where there's an ambiguous or invalid state. :) Not an issue, of course, if we just support `amount` and forget about shares.

cwang25: Sounds good  👍

test/MorphoActions.t.sol
cwang25: Imported SharesMathLib, and now it's comparing with the computed expected shares #

test/MorphoActions.t.sol
cwang25: abstracted shares input completely 👍

test/MorphoActions.t.sol
cwang25: Got it. Yeah usually I lean toward to keep consistent styles haha.  But I also open to create a refactor task to track (ofc it might not be in high priority but maybe worth picking it up sometime in the future)

test/MorphoVaultActions.t.sol
cwang25: Yeah I agree. 
I have removed this line as rethinking a bit this doesn't mean much to assert the share value. This function only tests if `morpho` interface can handle withdraw fund correctly. So I think just asserting wallet have the withdrawn fund is good enough.

src/MorphoScripts.sol
```diff
@@ -0,0 +1,182 @@
...
+                receiver: address(this)
+            });
+        }
+    }
+
+    // Helper function to convert MarketParams to bytes32 Id
+    // Reference: https://github.com/morpho-org/morpho-blue/blob/731e3f7ed97cf15f8fe00b86e4be5365eb3802ac/src/libraries/MarketParamsLib.sol
+    function marketId(MarketParams memory params) public pure returns (bytes32 marketParamsId) {
+        assembly ("memory-safe") {
+            marketParamsId := keccak256(params, 160)
```
fluffywaffles, line 136: why not `keccak256(abi.encode(params))`?

test/MorphoActions.t.sol
```diff
@@ -0,0 +1,165 @@
...
+        wallet.executeQuarkOperation(op, v, r, s);
+
+        assertEq(IERC20(USDC).balanceOf(address(wallet)), 200e6);
+        assertEq(IERC20(wstETH).balanceOf(address(wallet)), 5e18);
+        assertEq(IMorpho(morpho).position(marketId(marketParams), address(wallet)).collateral, 5e18);
+        assertApproxEqAbs(
+            IMorpho(morpho).position(marketId(marketParams), address(wallet)).borrowShares,
+            SharesMathLib.toSharesUp(200e6, totalBorrowAssets, totalBorrowShares),
+            1
+        );
```
fluffywaffles, line 89: nice, this is a lot more readable: 200USDC toSharesUp makes sense! :raised_hands:

fluffywaffles (approved):
looks great

src/MorphoScripts.sol
cwang25: this one works too `keccak256(abi.encode(params))`,  actually I was initially using this one happily haha.

But then I saw Morpho's implementation is using `assembly` 
So ended up sticking with their implementation 1:1 as I have put the `reference` comment above to avoid potential confusion in the future (maybe not much).

PR #70, opened 2024-09-04
Switch to Uniswap `SwapRouter02` from `SwapRouter`

kevincheng96:
We switch to Uniswap's `SwapRouter02` from `SwapRouter` because `SwapRouter` isn't supported on chains like Base. The routers are mostly the same, with `SwapRouter02` being newer and having support for Uniswap v2. However, their interfaces do differ slightly as the swap params for `SwapRouter02` no longer take a `deadline`. This is fine since a) recurring swaps don't need a deadline anyways and b) we can use the QuarkOperation's `expiry` to set the deadline for one-time swaps.

We have to manually import parts of `ISwapRouter02` because one of its dependencies [locks the compiler version](https://github.com/Uniswap/swap-router-contracts/issues/72) to `=0.7.6`, so the library cannot be directly used for higher Solidity versions without modifications.


foundry.toml
```diff
@@ -12,5 +12,8 @@ via_ir = true
 optimizer = true
 optimizer_runs = 100000000
 
+[fmt]
+ignore = ["src/vendor/**/*"]
+
```
kevincheng96, line 17: We want to ignore vendored contracts from being formatted to limit the amount of the diffs from the source.

banky (approved):
Cool, thanks!

fluffywaffles (approved):

PR #71, opened 2024-09-05
Move off of Compound node provider proxy

kevincheng96:
Similar change to https://github.com/compound-finance/quark/pull/206

### DIFF
--- a/.github/workflows/gas-snapshot.yml
+++ b/.github/workflows/gas-snapshot.yml
@@ -33,7 +33,7 @@ jobs:
             | grep -E '^test'   \
             | tee .gas-snapshot.new
         env:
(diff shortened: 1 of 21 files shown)


coburncoburn (approved):

cwang25:
Can you also share the MAINNET_RPC_URL value in 1Password as well :) 
So I can set it up in my machine and run `forge test`

kevincheng96:
> Can you also share the MAINNET_RPC_URL value in 1Password as well :) So I can set it up in my machine and run `forge test`

Any node URL works, like Infura or Alchemy. There are already log-in instructions in 1password that you can use to grab the keys.

PR #73, opened 2024-09-10
Morpho claim reward builder code (p4)

cwang25:


src/builder/QuarkBuilder.sol
```diff
@@ -1589,6 +1589,140 @@ contract QuarkBuilder {
...
+    struct MorphoRewardsClaimIntent {
+        uint256 blockTimestamp;
+        address claimer;
+        uint256 chainId;
+        address[] accounts;
+        uint256[] claimables;
+        address[] distributors;
+        address[] rewards;
+        bytes32[][] proofs;
+    }
```
fluffywaffles, line 1601: do we have an idea for how the client is going to obtain the data for this intent?

`claimables`, `accounts`, `distributors`, `rewards`, and `proofs` are a lot to manage.

it seems like a more ideal intent would be: "claim all my Morpho rewards, here's my `chainId`, `timestamp`, and the address of my wallet to use for the `claimer`". There's obviously a lot more than that here. How much of this can we abstract away? Where is the client going to figure out the rest of the data that we cannot embed in `QuarkBuilder` for them?

at a first blush, `accounts` we can derive for the user by enumerating their morpho positions from their chain accounts and taking an array of the addresses that have a morpho position. Then they don't have to pass that in.

`distributors` we can probably embed into builderpack, since that's the URD contract addresses.

the set of `rewards` tokens to claim -- also we should embed into builderpack.

`claimables` -- that's awkward, you have to specify amounts for each token? :thinking: 

`proofs` -- this one is gritty, IIRC you literally have to invoke an API to get a merkle root or something. :weary: 

---

basically, ignoring that `Morpho` rewards is this grotesquely abstract tries-to-be-everything-to-everyone system, what's the _actual_ user intent? and can we refine that to the minimum input?

because I don't think the client wants to specify _which_ rewards tokens, or the amounts, or from _where_; they just want to "claim all my rewards," and we should try to figure out the rest on their behalf. I think we can do that if we take an opinionated stance on things like (a) which URD contracts to care about (ones applicable to our vaults), (b) which rewards tokens to claim (tokens our wallet supports), and (c) how much to claim (always the maximum amount). stuff like that.

test/builder/QuarkBuilderMorphoClaimRewards.t.sol
```diff
@@ -0,0 +1,246 @@
...
+import {TransferActions} from "src/DeFiScripts.sol";
+import {MorphoInfo} from "src/builder/MorphoInfo.sol";
+import {MorphoRewardsActions} from "src/MorphoScripts.sol";
+import {List} from "src/builder/List.sol";
+import {Paycall} from "src/Paycall.sol";
+import {QuarkBuilder} from "src/builder/QuarkBuilder.sol";
+
+contract QuarkBuilderMorphoClaimRewardsTest is Test, QuarkBuilderTest {
+    // Mock morpho reward data to pass in
+    address[] mockDistributors =
```
fluffywaffles, line 21: this is not mock, is it? these are actual distributors. I would call it a fixture.

test/builder/QuarkBuilderMorphoClaimRewards.t.sol
```diff
@@ -0,0 +1,246 @@
...
+        [0x330eefa8a787552DC5cAd3C3cA644844B1E61Ddb, 0x330eefa8a787552DC5cAd3C3cA644844B1E61Ddb];
+    address[] mockAccounts = [address(0xa11ce), address(0xa11ce)];
+    address[] mockRewards = [usdc_(1), weth_(1)];
+    address[] mockInvalidRewards = [usdc_(1)];
+    uint256[] mockClaimables = [100e6, 2e18];
+    uint256[] mockClaimablesLessUSDC = [1e6, 2e18];
+    bytes32[][] mockProofs = [
```
fluffywaffles, line 28: similarly, this is a proofs fixture, right? not a mock.

fluffywaffles (commented):
I think we can do better on the intent here, and should try to think it through.

src/builder/QuarkBuilder.sol
cwang25: Hmm the problem will be builder code will never know what are those values.  Since those are only available to their API calls,  we need those values to be passed in somewhere. 

I don't think making client to call explicitly to morpho's api is good, even though their api is freely available to public).  But we might can let backend to load those values like portfolio snapshot and pass into builder code in a form like ChainAccountList.  But client would probably likely need to load those backend data to pass to buildercode just like portfolio snapshot.


For `distributors`, there is no guaranteed to be only a finite set of the addresses. So far my observation it is correlated with their reward programs.  If one market has new/re-new/special reward program, the api will probably return it as a new `item` in their response list with new `distributors`.

`rewards` yeah unfortunately that's required by Morpho input, and yes if we only allow certain tokens, we can embed those into code. 


But from a bright side, all `accounts` `distributors` `rewards` `claimables` and `proofs` are all provided from api response one-to-one.  So if we load the api response from either client or backend, there will be no work on our side to come out the values of those, we just simply pass those to contract call, and it should claim the reward successfully out of the box.

test/builder/QuarkBuilderMorphoClaimRewards.t.sol
cwang25: oh yeah sure,  will change to fixture.  👍 

was just using the name, but annoyingly solidity complains the name conflicts with function parameters name :( 
So I slapped `mock` lol

src/builder/QuarkBuilder.sol
fluffywaffles: mmmk, so everything comes from Morpho's API...

their rewards architecture is so convoluted and complicated.

perhaps the easiest path is if the client sends a "partial intent" to the backend (idk `POST /intent/morpho-claim-rewards`), and expect back the full intent that it can pass into quark builder.

it seems reasonable if we let the frontend _ask for an intent_ from the backend, but then process the intent locally in `QuarkBuilder` to retain control over the important bits of constructing quark operation(s).

fluffywaffles (commented):
Let's rebase / merge this and then can take another look.

test/builder/QuarkBuilderMorphoClaimRewards.t.sol
cwang25: fixed

fluffywaffles (approved):

PR #74, opened 2024-09-10
Small bug fix on potential underflow error in constructBridgeOperations

cwang25:
- Noticed in one clause it does direct subtraction, which will cause underflow error if the chain has less than intended amount to bridge.


.gas-snapshot
kevincheng96, line 1: We should probably just turn off gas snapshots for files in `src/builder`.

test/builder/QuarkBuilderCometWithdraw.t.sol
```diff
@@ -496,4 +496,19 @@ contract QuarkBuilderCometWithdrawTest is Test, QuarkBuilderTest {
...
         );
     }
+
+    function testCometWithdrawCostTooHigh() public {
+        PaymentInfo.PaymentMaxCost[] memory maxCosts = new PaymentInfo.PaymentMaxCost[](2);
+        maxCosts[0] = PaymentInfo.PaymentMaxCost({chainId: 1, amount: 5e6});
+        maxCosts[1] = PaymentInfo.PaymentMaxCost({chainId: 8453, amount: 5e6});
+        QuarkBuilder builder = new QuarkBuilder();
+
+        vm.expectRevert(abi.encodeWithSelector(Actions.NotEnoughFundsToBridge.selector, "usdc", 4e6, 4e6));
```
kevincheng96, line 506: Shouldn't this be "USDC"? (capitalized)

kevincheng96 (approved):

kevincheng96:
Good catch

test/builder/QuarkBuilderCometWithdraw.t.sol
cwang25: That's good question... haha. 
I agree.. it should be capitalized,  let me double check... looks like all existing tests is verifying it with lower cased 😮 ...

kevincheng96: Weird that the check is case insensitive 🤔

cwang25: oh wow... it's case sensitive...  hmm I think I need to dig a bit deeper to see why some tests seems still carrying side effect of `toLowerCase()` and some don't.

PR #75, opened 2024-09-10
Add isExactOut on one-time swaps

banky:
The one-time swap intent is used to generate the recurring intent, so we need to add `isExactOut` to the one-time swap so it can be propagated over


kevincheng96 (approved):
LGTM

Though looks like you may be running the formatter using a different foundry version. We fix the foundry version to `nightly-de33b6af53005037b463318d2628b5cfcaf39916` when running the formatter.

PR #81, opened 2024-09-16
Quark v2 changes

kevincheng96:
Updating the `quark-scripts` repo to depend on the latest version of Quark ([Quark v2](https://github.com/compound-finance/quark/pull/216)). The main changes in this PR are:

- Update `quark` submodule to the latest commit on the Quark v2 branch
- Bump up Solidity version to 0.8.27 and EVM version to `cancun`
- Update `QuarkOperation` struct to the newest version that has a `isReplayable` field
- Update `RecurringSwap.sol` to no longer need a `cancel` function and no longer need to call `allowReplay()`
- Bring in the `Cancel.sol` core script, used for cancelling replayable operations
- Update `QuarkBuilder` to use Quark v2; add the new fields `nonceSecret` and `totalPlays` to QuarkBuilder
- Optimize `QuarkBuilder` to get rid of stack too deep errors from introducing the new fields


src/builder/Actions.sol
```diff
@@ -593,7 +593,8 @@ library Actions {
 
         // Construct QuarkOperation
         IQuarkWallet.QuarkOperation memory quarkOperation = IQuarkWallet.QuarkOperation({
-            nonce: accountState.quarkNextNonce,
+            nonce: bytes32(uint256(accountState.quarkNextNonce)),
```
kevincheng96, line 596: There's a bunch of interface changes that need to be made in `QuarkBuilder` for the client to pass in a random nonce. I plan to do that as a separate PR.

src/builder/PaycallWrapper.sol
```diff
@@ -25,6 +25,7 @@ library PaycallWrapper {
 
         return IQuarkWallet.QuarkOperation({
             nonce: operation.nonce,
+            isReplayable: operation.isReplayable,
```
hayesgm, line 28: This is interesting, and I think correct. The Paycall operation is repalyable IFF what it wraps is replayable?

src/builder/QuarkOperationHelper.sol
```diff
@@ -101,6 +101,7 @@ library QuarkOperationHelper {
         IQuarkWallet.QuarkOperation memory lastQuarkOperation = quarkOperations[quarkOperations.length - 1];
         IQuarkWallet.QuarkOperation memory mergedQuarkOperation = IQuarkWallet.QuarkOperation({
             nonce: lastQuarkOperation.nonce,
+            isReplayable: lastQuarkOperation.isReplayable,
```
hayesgm, line 104: Same here, so basically if it's `wrap` and `replayableSupply` this is now `replayableWrapAndSupply`?

hayesgm (approved):
lgtm

src/builder/PaycallWrapper.sol
kevincheng96: Exactly. If it's a `RecurringSwap` using `Paycall`, then the `Paycall` itself has to be replayable.

src/builder/QuarkOperationHelper.sol
kevincheng96: Yeah, if the last operation is replayable, the merged operation should be replayable.

kevincheng96:
The latest commit (fc9d2094763fd27231b9770fdf7d3ba09e333509) rebases on main and applies the Quark v2 changes to the newest `MorphoClaimRewards` action. After the rebase, we were hitting stack too deep errors again, so the commit also includes some optimizations to the `QuarkBuilder` code.

PR #83, opened 2024-09-23
Support Quark v2 in QuarkBuilder

kevincheng96:
Adds `nonceSecret` and `replayCount` to QuarkBuilder. For replayable operations, the QuarkBuilder will also generate the nonce using the `nonceSecret` and `replayCount` (see `generateNonceFromSecret` function).

We also uncover and fix a bug where `0` is being used as the `nonce` for an operation when a nonce is not provided for an account by the client. Now, the code reverts if a `nonceSecret` is not given for an account. We buff up the unit tests to assert the right values are being used for the nonces, since nonces were being checked previously.

### DIFF
--- a/src/builder/Accounts.sol
+++ b/src/builder/Accounts.sol
@@ -8,21 +8,20 @@ import {PaymentInfo} from "./PaymentInfo.sol";
 import {TokenWrapper} from "./TokenWrapper.sol";
 
 library Accounts {
+    error QuarkSecretNotFound(address account);
+
     struct ChainAccounts {
         uint256 chainId;
-        QuarkState[] quarkStates;
+        QuarkSecret[] quarkSecrets;
         AssetPositions[] assetPositionsList;
         CometPositions[] cometPositions;
         MorphoPositions[] morphoPositions;
         MorphoVaultPositions[] morphoVaultPositions;
     }
 
-    // We map this to the Portfolio data structure that the client will already have.
-    // This includes fields that builder may not necessarily need, however it makes
-    // the client encoding that much simpler.
-    struct QuarkState {
+    struct QuarkSecret {
         address account;
-        uint96 quarkNextNonce;
+        bytes32 nonceSecret;
     }
 
     // Similarly, this is designed to intentionally reduce the encoding burden for the client
@@ -182,16 +181,17 @@ library Accounts {
         return findAssetPositions(assetAddress, chainAccounts.assetPositionsList);
     }
 
-    function findQuarkState(address account, Accounts.QuarkState[] memory quarkStates)
+    function findQuarkSecret(address account, Accounts.QuarkSecret[] memory quarkSecrets)
         internal
         pure
-        returns (Accounts.QuarkState memory state)
+        returns (Accounts.QuarkSecret memory)
     {
-        for (uint256 i = 0; i < quarkStates.length; ++i) {
-            if (quarkStates[i].account == account) {
-                return state = quarkStates[i];
+        for (uint256 i = 0; i < quarkSecrets.length; ++i) {
+            if (quarkSecrets[i].account == account) {
+                return quarkSecrets[i];
             }
         }
+        revert QuarkSecretNotFound(account);
     }
 
     function findChainAccountsWithPaymentInfo(
--- a/src/builder/Actions.sol
+++ b/src/builder/Actions.sol
@@ -58,8 +58,8 @@ library Actions {
     uint256 constant SWAP_EXPIRY_BUFFER = 3 days;
     uint256 constant TRANSFER_EXPIRY_BUFFER = 7 days;
 
-    /* replay counts */
-    uint256 constant RECURRING_SWAP_REPLAY_COUNT = 500;
+    /* total plays */
+    uint256 constant RECURRING_SWAP_TOTAL_PLAYS = 500;
 
     uint256 constant AVERAGE_BLOCK_TIME = 12 seconds;
     uint256 constant RECURRING_SWAP_MAX_SLIPPAGE = 1e17; // 1%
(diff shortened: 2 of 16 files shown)


test/builder/QuarkBuilderCometWithdraw.t.sol
```diff
@@ -253,23 +281,32 @@ contract QuarkBuilderCometWithdrawTest is Test, QuarkBuilderTest {
...
         QuarkBuilder.BuilderResult memory result = builder.cometWithdraw(
-            cometWithdraw_(8453, cometUsdc_(8453), "LINK", 5e18), chainAccountsList, paymentUsdc_(maxCosts)
+            // We need to set Bob as the withdrawer because only he has an account on chain 8453
+            cometWithdraw_({
+                chainId: 8453,
+                comet: cometUsdc_(8453),
+                assetSymbol: "LINK",
+                amount: 5e18,
+                withdrawer: address(0xb0b)
+            }),
```
kevincheng96, line 307: Example of the nonce bug in action. Previously, we were withdrawing using `0xa11ce` as the withdrawer on chain `8453`, even though a nonce for `0xa11ce` wasn't provided for chain 8453. Now, we have to set the withdrawer as `0xb0b` since only he has a nonce on chain 8453.

src/builder/Accounts.sol
```diff
@@ -182,16 +181,17 @@ library Accounts {
         return findAssetPositions(assetAddress, chainAccounts.assetPositionsList);
     }
 
-    function findQuarkState(address account, Accounts.QuarkState[] memory quarkStates)
+    function findQuarkSecret(address account, Accounts.QuarkSecret[] memory quarkSecrets)
```
hayesgm, line 184: Do we need to do it this way? Wouldn't just "next random" work?

src/builder/Actions.sol
```diff
@@ -58,6 +58,9 @@ library Actions {
     uint256 constant SWAP_EXPIRY_BUFFER = 3 days;
     uint256 constant TRANSFER_EXPIRY_BUFFER = 7 days;
 
+    /* replay counts */
+    uint256 constant RECURRING_SWAP_REPLAY_COUNT = 500;
```
hayesgm, line 62: I tend to think 500 is pretty high. We should just discuss as a product question generally

src/builder/Actions.sol
```diff
@@ -245,6 +256,11 @@ library Actions {
         address paymentToken;
         string paymentTokenSymbol;
         uint256 paymentMaxCost;
+        // The secret used to generate the hash chain for a replayable operation. For non-replayable
+        // operations, the `nonce` will be the `nonceSecret` (the hash chain has a length of 1)
+        bytes32 nonceSecret;
+        // The number of times an operation can be replayed. For non-replayable operations, this will be 0
+        uint256 replayCount;
```
hayesgm, line 263: On the backend, we've been more keen to use "totalPlays" which tends to map a little better since it doesn't have a discontinuity at 0.

src/builder/Actions.sol
```diff
@@ -245,6 +256,11 @@ library Actions {
         address paymentToken;
         string paymentTokenSymbol;
         uint256 paymentMaxCost;
+        // The secret used to generate the hash chain for a replayable operation. For non-replayable
```
hayesgm, line 259: Should we return `nonce` and `nonceSecret`? Generally we've been on "verify" mode on the client (as opposed to "generate"). That is, we run a check that `hash_n(input.nonceSecret, input.totalPlays) == input.nonce`

src/builder/Actions.sol
```diff
@@ -1487,6 +1544,18 @@ library Actions {
         return result;
     }
 
+    function generateNonceFromSecret(bytes32 secret, uint256 replayCount) internal pure returns (bytes32) {
+        assembly ("memory-safe") {
```
hayesgm, line 1548: cool-zone

test/builder/lib/ReplayableHelper.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity ^0.8.27;
+
+library ReplayableHelper {
+    function generateNonceFromSecret(bytes32 secret, uint256 replayCount) internal pure returns (bytes32) {
```
hayesgm, line 5: Any reason not to use the optimized version above?

src/builder/Accounts.sol
kevincheng96: Synced offline on this. I don't think we resolved this discussion yet, but here's my response:

I considered specifying `nonceSecrets` on a non-account basis, but I decided to keep it account-based for a few reasons.

1. It conforms to the current interface that the client is passing in.
2. It reduces the complexity of the QuarkBuilder having to pick out `nonceSecrets` and keep track of which ones are already assigned to which operations.

src/builder/Actions.sol
kevincheng96: Should be easy to change this when we want to.

src/builder/Actions.sol
kevincheng96: The actual `nonce` is on the `QuarkOperation`, so I don't think we need to put it here.

src/builder/Actions.sol
kevincheng96: Adjusted to `totalPlays` to be consistent. Both are fine imo.

test/builder/lib/ReplayableHelper.sol
kevincheng96: Mainly just because we don't need the optimized version here to save us from stack too deep. Also I thought it's kinda nice to have a more readable version so we can sanity check the optimized version in our tests with a different implementation.

src/builder/Accounts.sol
hayesgm: TBF, if we needed to do two operations on the same chain, this strategy wouldn't work? E.g. let's say we wanted to make two delayed operations on the same chain.

hayesgm (approved):

src/builder/Accounts.sol
kevincheng96: If we have two operations on the same chain for the same account, we wrap those in a multi quark operation. So having a single `nonceSecret` for an account on each chain is guaranteed to be sufficient.

PR #84, opened 2024-09-25
Add EIP-712 Change for Quark V2

hayesgm:
This patch adds an EIP-712 change for Quark V2, which adds an `isReplayable` field in the typehash of the EIP-712 struct.


kevincheng96 (approved):
Good catch

PR #90, opened 2024-10-17
CS5.7 - Prevent executing recurring swaps for missed intervals

kevincheng96:
Previously, if a recurring swap interval was missed in the past, a swap could be repeatedly executed for each missed interval. This change makes it so that swaps can no longer be executed for missed intervals.


src/RecurringSwap.sol
```diff
@@ -105,8 +105,12 @@ contract RecurringSwap is QuarkScript {
             revert SwapWindowNotOpen(nextSwapTime, block.timestamp);
         }
 
-        // Update nextSwapTime
-        write(hashedConfig, bytes32(nextSwapTime + config.interval));
+        // Update the next swap time, ensuring that it is some time in the future
+        uint256 updatedNextSwapTime = nextSwapTime;
+        while (updatedNextSwapTime <= block.timestamp) {
+            updatedNextSwapTime += config.interval;
```
kevincheng96, line 111: The real question is if we should alter the logic to be `nextSwapTime = block.timestamp + config.interval`. The problem with that approach is it could lead to time drift, where a "Swap every Monday" slowly shifts to a "Swap every Tuesday" over time.

hayesgm: Isn’t there a world where we check the window and mark that window as closed? E.g. if it’s Mondays starting Jan 1, then after the first Monday we say “next window=1” implying Jan 8. And if we miss a few we simply find the window we are now in and say “next window = current window + 1”?

fluffywaffles (commented):
Left some notes in `#recurring-buys` out-of-band from GitHub. Mostly curious if we really want "7 day windows" vs. "roughly every 7 days," where the error bars are tighter and arguably the user intent is more accurately represented.

test/RecurringSwap.t.sol
```diff
@@ -542,6 +542,58 @@ contract RecurringSwapTest is Test {
...
+            isExactOut: true
+        });
+        (QuarkWallet.QuarkOperation memory op, bytes32[] memory submissionTokens) = new QuarkOperationHelper()
+            .newReplayableOpWithCalldata(
+            aliceWallet,
+            recurringSwap,
+            abi.encodeWithSelector(RecurringSwap.swap.selector, swapConfig),
+            ScriptType.ScriptAddress,
+            2
+        );
```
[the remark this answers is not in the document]
kevincheng96: Yeah this is `forge fmt`'s doing. Unfortunately adding the line break doesn't actually affect how it formats.

src/RecurringSwap.sol
kevincheng96: Discussed offline. We decided to add a `length` to the swap window. Basically, the swap is executable any time between the `swapWindowStartTime + swapWindowLength`. The `swapWindowStartTime` shifts in increments of `swapWindowInterval`.

src/RecurringSwap.sol
```diff
@@ -93,20 +102,29 @@ contract RecurringSwap is QuarkScript {
...
-            revert SwapWindowNotOpen(nextSwapTime, block.timestamp);
+        // Check that swap window is open
+        if (block.timestamp < nextWindowStart) {
+            revert SwapWindowNotOpen(nextWindowStart, config.swapWindow.length, block.timestamp);
+        }
+
+        // Find the last window start time and the next window start time
+        uint256 completedIntervals = (block.timestamp - config.swapWindow.startTime) / config.swapWindow.interval;
+        uint256 lastWindowStart = config.swapWindow.startTime + (completedIntervals * config.swapWindow.interval);
+        uint256 updatedNextWindowStart = lastWindowStart + config.swapWindow.interval;
```
kevincheng96, line 120: I updated this logic to use math instead of a while loop.

src/RecurringSwap.sol
```diff
@@ -93,20 +102,29 @@ contract RecurringSwap is QuarkScript {
         }
 
         bytes32 hashedConfig = _hashConfig(config);
```
hayesgm, line 104: Why do we use hashConfig here and not just isolated storage since it's based on script nonce?

src/RecurringSwap.sol
```diff
@@ -93,20 +102,29 @@ contract RecurringSwap is QuarkScript {
...
-        uint256 nextSwapTime;
+        uint256 nextWindowStart;
         if (read(hashedConfig) == 0) {
-            nextSwapTime = config.startTime;
+            nextWindowStart = config.swapWindow.startTime;
         } else {
-            nextSwapTime = uint256(read(hashedConfig));
+            nextWindowStart = uint256(read(hashedConfig));
```
hayesgm, line 109: It might just be prudent to say this is "completedIntervals" or "currentIntervalNumber" since it's a _bit_ more confusing to store a timestamp versus a simple small integer.

hayesgm: E.g. then a lot of the code below is basically "is currentInterval > lastInterval" and then "block.timestamp > currentInterval * intervalLength && block.timestamp <= currentInterval * intervalLength + windowLength", which is pretty close to what you have, I guess

hayesgm (approved):
lgtm

src/RecurringSwap.sol
kevincheng96: It's because a nonce can have multiple recurring swap orders. The hashConfig provides a level of isolation, on top of the existing nonce-isolation, if a user signs multiple orders for the same nonce. See this [test](https://github.com/compound-finance/quark-scripts/blob/ff0050165f462ec50bbd7d38335869adac30306d/test/RecurringSwap.t.sol#L443) for example.

src/RecurringSwap.sol
kevincheng96: Yeah, that could also work. I'll play around with this alternative implementation and start a new PR if it looks good.

PR #91, opened 2024-10-18
Update to latest Quark version

kevincheng96:
Updates the Quark submodule to the latest [commit](https://github.com/compound-finance/quark/commit/4e5ba95ac5118cca5456d41c8ddd9415eb443530) in the Quark repo , which notably includes the changes to:

1. Move from `v,r,s` to `bytes` signatures
2. Switch from `CALLCODE` to `DELEGATECALL` 
3. Remove the `onlyWallet` reentrancy guard in favor of the `nonReentrant` guard.


src/RecurringSwap.sol
```diff
@@ -132,7 +132,7 @@ contract RecurringSwap is QuarkScript {
 
         // Emit the swap event
         emit SwapExecuted(
-            msg.sender,
+            address(this),
```
kevincheng96, line 135: This change is due to switching from `CALLCODE` to `DELEGATECALL`.

hayesgm (commented):
Looks good only question is: v isn’t required to only be one byte, right? Are these EIP-15( signatures?

kevincheng96:
> Looks good only question is: v isn’t required to only be one byte, right? Are these EIP-15( signatures?

Yeah, the tests currently only use EIP-155 signatures, so `v` is guaranteed to be one bytes. If we test other signature schemes like BLS, we will need a new signature decoder since the signatures will have different lengths as well.

hayesgm (approved):

PR #92, opened 2024-10-22
CS5.1 - Disallow callbacks at the start of a script's callback function

kevincheng96:
Brings over the exact same [fixes](https://github.com/compound-finance/quark/pull/229) to the same files from the `quark` repo.


hayesgm (approved):
Thanks

banky (approved):

PR #94, opened 2024-10-24
Quark script to bridge using Across

kevincheng96:
Implements a Quark script for bridging assets across chains using Across.

### DIFF
--- a/src/AcrossScripts.sol
+++ b/src/AcrossScripts.sol
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity 0.8.27;
+
+import {IERC20} from "openzeppelin/token/ERC20/IERC20.sol";
+import {SafeERC20} from "openzeppelin/token/ERC20/utils/SafeERC20.sol";
+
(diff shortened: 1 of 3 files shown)


src/AcrossScripts.sol
```diff
@@ -0,0 +1,78 @@
...
+        address outputToken,
+        uint256 inputAmount,
+        uint256 outputAmount,
+        uint256 destinationChainId,
+        address exclusiveRelayer,
+        uint32 quoteTimestamp,
+        uint32 fillDeadline,
+        uint32 exclusivityDeadline,
+        bytes calldata message,
+        bool useNativeToken
```
kevincheng96, line 59: This is a big list of params. Some of the ones that we could consider setting default values for are: exclusiveRelayer, fillDeadline, exclusivityDeadline, message.

I decided for now to keep them for flexibility.

kevincheng96: We could also add a constructor that takes in the `spokePool` address, since that is supposed to be constant on each chain.

coburncoburn (approved):

PR #96, opened 2024-10-25
Add oracle staleness checks to RecurringSwap script

kevincheng96:
This adds a last-resort sanity check to ensure that oracles have been updated within the last 24 hours. I also add a `MockPriceFeed` to our tests so we can better test bad prices and stale prices.

### DIFF
--- a/test/RecurringSwap.t.sol
+++ b/test/RecurringSwap.t.sol
@@ -14,10 +14,10 @@ import {QuarkMinimalProxy} from "quark-proxy/src/QuarkMinimalProxy.sol";
 import {Cancel} from "src/Cancel.sol";
 import {RecurringSwap} from "src/RecurringSwap.sol";
 
-import {YulHelper} from "./lib/YulHelper.sol";
-import {SignatureHelper} from "./lib/SignatureHelper.sol";
-import {QuarkOperationHelper, ScriptType} from "./lib/QuarkOperationHelper.sol";
-import {MockPriceFeed} from "./mocks/MockPriceFeed.sol";
+import {YulHelper} from "test/lib/YulHelper.sol";
+import {SignatureHelper} from "test/lib/SignatureHelper.sol";
+import {QuarkOperationHelper, ScriptType} from "test/lib/QuarkOperationHelper.sol";
+import {MockPriceFeed} from "test/mocks/MockPriceFeed.sol";
 
 import {AggregatorV3Interface} from "src/vendor/chainlink/AggregatorV3Interface.sol";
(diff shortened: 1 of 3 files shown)


test/RecurringSwap.t.sol
```diff
@@ -17,7 +17,9 @@ import {RecurringSwap} from "src/RecurringSwap.sol";
 import {YulHelper} from "./lib/YulHelper.sol";
 import {SignatureHelper} from "./lib/SignatureHelper.sol";
 import {QuarkOperationHelper, ScriptType} from "./lib/QuarkOperationHelper.sol";
+import {MockPriceFeed} from "./mocks/MockPriceFeed.sol";
 
+import {AggregatorV3Interface} from "src/vendor/chainlink/AggregatorV3Interface.sol";
```
cwang25, line 22: Should we use absolute path for the rest as well? 😄

cwang25 (approved):
Overall lgtm! 

But I'm wondering if we should consider a potential of interacting with some hard-coded pricefeed (like comet's constant pricefeed for weth/eth) that might not ever have "fresh" updatedAt timestamp?

kevincheng96:
> But I'm wondering if we want to consider a potential of interacting with some hard-coded pricefeed (like comet's constant pricefeed for weth/eth) that might not ever have "fresh" updatedAt timestamp?

That's a great point, though Comet's [constant price feed](https://etherscan.io/address/0xD72ac1bCE9177CFe7aEb5d0516a38c88a64cE0AB) always returns the `block.timestamp` as the `updatedAt` value, so this script won't have any issues using it. I think it's safe to assume that any custom price feed should be returning a non-stale `updatedAt` value, otherwise something is wrong with the price feed itself.

cwang25:
> > But I'm wondering if we want to consider a potential of interacting with some hard-coded pricefeed (like comet's constant pricefeed for weth/eth) that might not ever have "fresh" updatedAt timestamp?
> 
> That's a great point, though Comet's [constant price feed](https://etherscan.io/address/0xD72ac1bCE9177CFe7aEb5d0516a38c88a64cE0AB) always returns the `block.timestamp` as the `updatedAt` value, so this script won't have any issues using it. I think it's safe to assume that any custom price feed should be returning a non-stale `updatedAt` value, otherwise something is wrong with the price feed itself.

I agree, sounds good 👍😆
Also browsed through some pricefeeds on chain link can also confirm that most feeds heartbeat is less than 1 day.

test/RecurringSwap.t.sol
kevincheng96: Done

PR #97, opened 2024-10-28
add Arbitrum (mainnet+sepolia) addresses

scott-silver:


src/builder/BridgeRoutes.sol
```diff
@@ -36,6 +37,8 @@ library CCTP {
         chains[2] = CCTPChain({chainId: 11155111, domainId: 0, bridge: 0x9f3B8679c73C2Fef8b59B4f3444d4e156fb70AA5});
         // Base Sepolia
         chains[3] = CCTPChain({chainId: 84532, domainId: 6, bridge: 0x9f3B8679c73C2Fef8b59B4f3444d4e156fb70AA5});
+        // Arbitrum Sepolia
+        chains[4] = CCTPChain({chainId: 421614, domainId: 3, bridge: 0x9f3B8679c73C2Fef8b59B4f3444d4e156fb70AA5});
```
scott-silver, line 41: source: https://developers.circle.com/stablecoins/evm-smart-contracts

src/builder/UniswapRouter.sol
```diff
@@ -21,6 +21,8 @@ library UniswapRouter {
         chains[2] = RouterChain({chainId: 11155111, router: 0x3bFA4769FB09eefC5a80d6E87c3B9C650f7Ae48E});
         // Base Sepolia
         chains[3] = RouterChain({chainId: 84532, router: 0x94cC0AaC535CCDB3C01d6787D6413C739ae12bc4});
+        // Arbitrum Sepolia
+        chains[4] = RouterChain({chainId: 421614, router: 0x101F443B4d1b059569D643917553c771E1b9663E});
```
scott-silver, line 25: source: https://docs.uniswap.org/contracts/v3/reference/deployments/arbitrum-deployments

scott-silver:
Not sure if we need to update this contract as well (for Arbitrum Sepolia):

https://github.com/compound-finance/quark-scripts/blob/main/src/lib/UniswapFactoryAddress.sol

kevincheng96:
> Not sure if we need to update this contract as well (for Arbitrum Sepolia):
> 
> https://github.com/compound-finance/quark-scripts/blob/main/src/lib/UniswapFactoryAddress.sol

Probably would be a good idea to do that as well. Though none of the contracts that currently use that library are deployed and being used right now.

kevincheng96 (approved):
Double checked the addresses and they all look good 👍

coburncoburn (approved):

PR #98, opened 2024-10-30
Use FFI to perform paycall multi-step simulation

banky:
Currently, the logic for multi-step simulation for paycall transactions is all done on the client side. This PR aims to move this logic into QuarkBuilder by making a call to a simulation ffi. ie. when calling the builder with an intent which uses paycall, the builder first simulates the transaction to get the estimated max payment cost and then uses the estimate for building up the transaction with paycall

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -6,12 +6,14 @@ solc = "0.8.27"
 evm_version = "cancun"
 bytecode_hash = "none"
 cbor_metadata = false
+# Ignore the tstore warning
+ignored_error_codes = [2394]
 
 libs = ["./lib"]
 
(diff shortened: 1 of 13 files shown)


foundry.toml
```diff
@@ -6,12 +6,14 @@ solc = "0.8.27"
...
 cbor_metadata = false
+# Ignore the tstore warning
+ignored_error_codes = [2394]
 
 libs = ["./lib"]
 
 [profile.ir]
 via_ir = true
-optimizer = true
+optimizer = false
```
banky, line 16: This should not go into main, just using locally to make the compiler a bit faster

PR #100, opened 2024-11-04
Add Across bridge support to the QuarkBuilder

kevincheng96:
This PR adds support for bridging assets using Across to the QuarkBuilder. This is the first time we are using an FFI in the QuarkBuilder, so any downstream clients using this version of the builder will need to implement the FFI at the reserved address (`address constant ACROSS_FFI_ADDRESS = address(0xFF1000)`).

### DIFF
--- a/src/builder/FFI.sol
+++ b/src/builder/FFI.sol
@@ -8,8 +8,9 @@ import {IAcrossFFI} from "src/interfaces/IAcrossFFI.sol";
  * @notice Defines the addresses of reserved FFIs and methods for calling them
  */
 library FFI {
-    /* FFI Addresses (starts from 0xFF1000, FFI with 100 reserved addresses) */
-    address constant ACROSS_FFI_ADDRESS = address(0xFF1000);
+    /// FFI Addresses (starts from 0xFF1000, FFI with 1000 reserved addresses)
+    /// 0xFF1000-0xFF1009 are reserved for framework-level FFIs like console log
+    address constant ACROSS_FFI_ADDRESS = address(0xFF1010);
 
     function requestAcrossQuote(
         address inputToken,
--- a/src/builder/FFI.sol
+++ b/src/builder/FFI.sol
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity ^0.8.27;
+
+import {IAcrossFFI} from "src/interfaces/IAcrossFFI.sol";
+
+/**
+ * @title Foreign Function Interface (FFI) Helper
+ * @notice Defines the addresses of reserved FFIs and methods for calling them
+ */
+library FFI {
+    /* FFI Addresses (starts from 0xFF1000, FFI with 100 reserved addresses) */
+    address constant ACROSS_FFI_ADDRESS = address(0xFF1000);
+
+    function requestAcrossQuote(
+        address inputToken,
+        address outputToken,
+        uint256 srcChain,
+        uint256 dstChain,
+        uint256 amount
+    ) internal pure returns (uint256 gasFee, uint256 variableFeePct) {
+        // Make FFI call to fetch a quote from Across API
+        return IAcrossFFI(ACROSS_FFI_ADDRESS).requestAcrossQuote(inputToken, outputToken, srcChain, dstChain, amount);
+    }
+}
--- a/src/builder/QuarkBuilderBase.sol
+++ b/src/builder/QuarkBuilderBase.sol
@@ -16,6 +16,7 @@ import {PaymentInfo} from "src/builder/PaymentInfo.sol";
 import {TokenWrapper} from "src/builder/TokenWrapper.sol";
 import {QuarkOperationHelper} from "src/builder/QuarkOperationHelper.sol";
 import {List} from "src/builder/List.sol";
+import {HashMap} from "src/builder/HashMap.sol";
 
 contract QuarkBuilderBase {
     /* ===== Output Types ===== */
@@ -36,7 +37,7 @@ contract QuarkBuilderBase {
 
     /* ===== Constants ===== */
 
-    string constant VERSION = "0.1.2";
+    string constant VERSION = "0.1.3";
 
     /* ===== Custom Errors ===== */
 
@@ -86,7 +87,7 @@ contract QuarkBuilderBase {
         Actions.Action memory action
     )
         internal
-        pure
+        view
         returns (IQuarkWallet.QuarkOperation[] memory quarkOperationsArray, Actions.Action[] memory actionsArray)
     {
         // Sanity check on ActionIntent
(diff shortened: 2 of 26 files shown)


src/interfaces/IFFI.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity 0.8.27;
+
+/// @dev Interface for foreign function interface (FFI) contracts
+interface IFFI {
```
hayesgm, line 5: Curious why this isn't specific to Across, e.g. `IFFIAcross`, since different contracts will have different addresses, right?

src/builder/actions/Actions.sol
```diff
@@ -676,6 +687,100 @@ library Actions {
...
+            Accounts.findAssetPositions(bridge.assetSymbol, dstChainAccounts.assetPositionsList);
+
+        Accounts.QuarkSecret memory accountSecret =
+            Accounts.findQuarkSecret(bridge.sender, srcChainAccounts.quarkSecrets);
+
+        bytes[] memory scriptSources = new bytes[](1);
+        scriptSources[0] = Across.bridgeScriptSource();
+
+        // Make FFI call to fetch a quote from Across API
+        bytes memory ffiCalldata = abi.encodeCall(
```
hayesgm, line 714: In `FFI.sol`, I wonder if we should have something akin to `IAcrossFFI` as well as `AcrossFFI` and just call `AcrossFFI.requestQuote(..)` and then this logic (and the FFI contract address) moves there. It would probably be a slightly neater way to bundle this from the code-caller's perspective.

src/builder/actions/Actions.sol
```diff
@@ -676,6 +687,100 @@ library Actions {
...
+            IFFI.requestAcrossQuote,
+            (
+                srcAssetPositions.asset,
+                dstAssetPositions.asset,
+                bridge.srcChainId,
+                bridge.destinationChainId,
+                bridge.amount
+            )
+        );
+        (bool success, bytes memory result) = ACROSS_FFI_ADDRESS.staticcall(ffiCalldata);
```
hayesgm, line 724: Do we need staticcall here? I wonder if we mark the function as `pure` or `view` if we can get Solidity to do `STATICCALL` without resorting to low-level calls?

hayesgm: E.g. https://docs.soliditylang.org/en/latest/contracts.html#view-functions

> View Functions[](https://docs.soliditylang.org/en/latest/contracts.html#view-functions)
> Functions can be declared view in which case they promise not to modify the state.
> **Note**
> If the compiler’s EVM target is Byzantium or newer (default) the opcode STATICCALL is used when view functions are called, which enforces the state to stay unmodified as part of the EVM execution. For library view functions DELEGATECALL is used, because there is no combined DELEGATECALL and STATICCALL. This means library view functions do not have run-time checks that prevent state modifications. This should not impact security negatively because library code is usually known at compile-time and the static checker performs compile-time checks.

test/builder/BridgingLogic.t.sol
```diff
@@ -0,0 +1,208 @@
...
+import {Quotecall} from "src/Quotecall.sol";
+import {YulHelper} from "test/lib/YulHelper.sol";
+
+import {AcrossFFI} from "test/builder/mocks/AcrossFFI.sol";
+
+contract BridgingLogicTest is Test, QuarkBuilderTest {
+    function setUp() public {
+        // Deploy mock FFI for calling Across API
+        AcrossFFI mockFFI = new AcrossFFI();
+        vm.etch(Actions.ACROSS_FFI_ADDRESS, address(mockFFI).code);
```
hayesgm, line 33: Well, I was about to write a long rant about how to test this and you nailed it here. Awesome.

hayesgm (approved):
looks awesome awesome. biggest Q is simply how to make a clean interface for FFI. Basically, I think we should add `Console.log` to FFI, and so we should make them easy to use as it'll come up a lot more than just here. Also, maybe start Bridge at `0xFF1010` or so, so we can reserve about 16 addresses for "core" features (e.g. console logging)

coburncoburn (approved):

src/builder/QuarkBuilderBase.sol
```diff
@@ -36,7 +37,7 @@ contract QuarkBuilderBase {
 
     /* ===== Constants ===== */
 
-    string constant VERSION = "0.1.2";
+    string constant VERSION = "0.1.3";
```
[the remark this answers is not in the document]
kevincheng96: Updated

src/builder/actions/Actions.sol
kevincheng96: Good suggestion! I moved this logic out into `FFI.sol`. The FFI addresses are also defined there now.

src/builder/actions/Actions.sol
kevincheng96: Oh interesting. I didn't know that the compiler uses STATICCALL under the hood for `pure/view` functions. Ok, I updated the code to no longer use `staticcall` directly so we can continue defining these functions as `pure`.

src/interfaces/IFFI.sol
kevincheng96: My original intention was to have a single interface file where all the FFIs are defined. But now that we have `FFI.sol`, there's not much benefit to keep all the FFIs in one interface. I updated this to be `IAcrossFFI.sol`

src/builder/FFI.sol
```diff
@@ -0,0 +1,24 @@
...
+
+import {IAcrossFFI} from "src/interfaces/IAcrossFFI.sol";
+
+/**
+ * @title Foreign Function Interface (FFI) Helper
+ * @notice Defines the addresses of reserved FFIs and methods for calling them
+ */
+library FFI {
+    /* FFI Addresses (starts from 0xFF1000, FFI with 100 reserved addresses) */
+    address constant ACROSS_FFI_ADDRESS = address(0xFF1000);
```
hayesgm, line 12: If we reserve 0x100 from `0xFF1000` then this should be 0xFF1100`, right? Unless we're reserving 0x1000 from from `0xFF0000` then that would align, but I don't think this maps with the comment.

src/builder/FFI.sol
```diff
@@ -0,0 +1,24 @@
...
+
+    function requestAcrossQuote(
+        address inputToken,
+        address outputToken,
+        uint256 srcChain,
+        uint256 dstChain,
+        uint256 amount
+    ) internal pure returns (uint256 gasFee, uint256 variableFeePct) {
```
hayesgm, line 20: curious if this is better to be a `view`. In better languages, improperly making something `pure` might allow optimizations to memoize the value, but given we know Solidity doesn't do that, I'm not too concern if we keep this `pure`, even knowing it has side-effects.

src/builder/FFI.sol
kevincheng96: Updated the comment to reflect that there are 1000 reserved addresses

src/builder/FFI.sol
kevincheng96: Yeah, it shouldn't matter in Solidity, so no strong preference here

PR #103, opened 2024-11-14
Bridge via Hinting

hayesgm:
This patch simply reverses the order s.t. we bridge via Across if possible, then CCTP otherwise based on a user hint (`preferAcross`).

Also, we move first-class support for console logs. Note: (a) we'll need to implement console logs in Eth.swift, and (b) we should just update `forge-std` and then use `import {console} from "forge-std/console.sol";` but upgrading forge-std causes stack too deep issues.


src/builder/actions/Actions.sol
```diff
@@ -610,9 +612,11 @@ library Actions {
         pure
         returns (IQuarkWallet.QuarkOperation memory, Action memory)
     {
-        if (CCTP.canBridge(bridge.srcChainId, bridge.destinationChainId, bridge.assetSymbol)) {
+        if (Across.canBridge(bridge.srcChainId, bridge.destinationChainId, bridge.assetSymbol)) {
+            console.log("Bridging via Across", bridge.assetSymbol);
             return bridgeUSDC(bridge, payment, useQuotecall);
-        } else if (Across.canBridge(bridge.srcChainId, bridge.destinationChainId, bridge.assetSymbol)) {
+        } else if (CCTP.canBridge(bridge.srcChainId, bridge.destinationChainId, bridge.assetSymbol)) {
+            console.log("Bridging via CCTP", bridge.assetSymbol);
```
kevincheng96, line 619: You'll need to switch the `bridgeUSDC` and `bridgeAcross` as well

src/builder/actions/Actions.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: BSD-3-Clause
 pragma solidity ^0.8.27;
 
+import {console} from "../console.sol";
```
kevincheng96, line 4: nit: we've been preferring absolute paths over relative paths for imports

kevincheng96, line 4: Can't you just import it from Foundry directly? via `import {console} from "forge-std/console.sol"`

test/builder/QuarkBuilderTransfer.t.sol
```diff
@@ -424,6 +455,163 @@ contract QuarkBuilderTransferTest is Test, QuarkBuilderTest {
         assertNotEq(result.eip712Data.hashStruct, hex"", "non-empty hashStruct");
     }
 
+    function testSimpleBridgeWithAcrossTransferSucceeds() public {
```
kevincheng96, line 458: I think a better place to put this would be in `BridgeLogic.t.sol` so we have a unified place to find the bridging tests.

kevincheng96 (approved):
Simple but effective. Nice work

src/builder/actions/Actions.sol
hayesgm: Yes, I could, but then I get a stack too deep error. Go figure :(

PR #105, opened 2024-11-18
Add "wrap/unwrap all" functions; always wrap/unwrap all in QuarkBuilder

kevincheng96:
~~We start using the new "wrap up to" scripts for ETH/WETH in the QuarkBuilder. In the case that ETH/WETh is bridged, we also need to always attempt to wrap/unwrap ETH/WETH on the destination chain because the QuarkBuilder cannot know ahead of time if Across will be sending ETH or WETH to the wallet.~~

We implement new `wrap/unwrap all` functions for ETH/WETH and stETH/wstETH. We also update the QuarkBuilder to always wrap/unwrap all (except for WETH, where we use "unwrap up to" because we prefer to preserve a user's WETH balance).

### DIFF
--- a/src/WrapperScripts.sol
+++ b/src/WrapperScripts.sol
@@ -10,6 +10,13 @@ contract WrapperActions {
         IWETH(weth).deposit{value: amount}();
     }
 
+    function wrapETHUpTo(address weth, uint256 targetAmount) external payable {
+        uint256 currentBalance = IERC20(weth).balanceOf(address(this));
+        if (currentBalance < targetAmount) {
+            IWETH(weth).deposit{value: targetAmount - currentBalance}();
+        }
+    }
+
     function wrapAllETH(address weth) external payable {
         uint256 ethBalance = address(this).balance;
         if (ethBalance > 0) {
@@ -19,6 +26,13 @@ contract WrapperActions {
 
     function unwrapWETH(address weth, uint256 amount) external {
         IWETH(weth).withdraw(amount);
+    }
+
+    function unwrapWETHUpTo(address weth, uint256 targetAmount) external {
+        uint256 currentBalance = address(this).balance;
+        if (currentBalance < targetAmount) {
+            IWETH(weth).withdraw(targetAmount - currentBalance);
+        }
     }
 
     function unwrapAllWETH(address weth) external payable {
--- a/src/WrapperScripts.sol
+++ b/src/WrapperScripts.sol
@@ -10,10 +10,10 @@ contract WrapperActions {
         IWETH(weth).deposit{value: amount}();
     }
 
-    function wrapETHUpTo(address weth, uint256 targetAmount) external payable {
-        uint256 currentBalance = IERC20(weth).balanceOf(address(this));
-        if (currentBalance < targetAmount) {
-            IWETH(weth).deposit{value: targetAmount - currentBalance}();
+    function wrapAllETH(address weth) external payable {
+        uint256 ethBalance = address(this).balance;
+        if (ethBalance > 0) {
+            IWETH(weth).deposit{value: ethBalance}();
         }
     }
 
@@ -21,10 +21,10 @@ contract WrapperActions {
         IWETH(weth).withdraw(amount);
     }
 
-    function unwrapWETHUpTo(address weth, uint256 targetAmount) external {
-        uint256 currentBalance = address(this).balance;
-        if (currentBalance < targetAmount) {
-            IWETH(weth).withdraw(targetAmount - currentBalance);
+    function unwrapAllWETH(address weth) external payable {
+        uint256 wethBalance = IERC20(weth).balanceOf(address(this));
+        if (wethBalance > 0) {
+            IWETH(weth).withdraw(wethBalance);
         }
     }
 
@@ -33,7 +33,22 @@ contract WrapperActions {
         IWstETH(wstETH).wrap(amount);
     }
 
+    function wrapAllLidoStETH(address wstETH, address stETH) external payable {
+        uint256 stETHBalance = IERC20(stETH).balanceOf(address(this));
+        if (stETHBalance > 0) {
+            IERC20(stETH).approve(wstETH, stETHBalance);
+            IWstETH(wstETH).wrap(stETHBalance);
+        }
+    }
+
     function unwrapLidoWstETH(address wstETH, uint256 amount) external {
         IWstETH(wstETH).unwrap(amount);
+    }
+
+    function unwrapAllLidoWstETH(address wstETH) external {
+        uint256 wstETHBalance = IERC20(wstETH).balanceOf(address(this));
+        if (wstETHBalance > 0) {
+            IWstETH(wstETH).unwrap(wstETHBalance);
+        }
     }
 }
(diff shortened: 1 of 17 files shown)


src/builder/TokenWrapper.sol
```diff
@@ -128,7 +128,7 @@ library TokenWrapper {
     {
         if (Strings.stringEqIgnoreCase(tokenSymbol, "ETH")) {
             return abi.encodeWithSelector(
-                WrapperActions.wrapETH.selector, getKnownWrapperTokenPair(chainId, tokenSymbol).wrapper, amount
+                WrapperActions.wrapETHUpTo.selector, getKnownWrapperTokenPair(chainId, tokenSymbol).wrapper, amount
```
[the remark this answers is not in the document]
kevincheng96: Updated it to only use `wrap up to` when ETH/WETH is bridged using Across.

cwang25 (approved):

src/builder/BridgeRoutes.sol
```diff
@@ -196,4 +197,18 @@ library Across {
...
     }
+
+    // Returns whether or not an asset is bridged non-deterministically. This applies to WETH/ETH, where Across will send either ETH or WETH
+    // to the target address depending on if address is an EOA or contract.
+    function isNonDeterministicBridgeAction(HashMap.Map memory assetsBridged, string memory assetSymbol)
+        internal
+        pure
+        returns (bool)
+    {
+        uint256 bridgedAmount = HashMap.contains(assetsBridged, abi.encode(assetSymbol))
```
hayesgm, line 208: Does the hashmap return zero for non-existent values or does it revert? Since otherwise this could just default to zero as a valid code path?

src/builder/BridgeRoutes.sol
```diff
@@ -196,4 +197,18 @@ library Across {
...
+            ? HashMap.getUint256(assetsBridged, abi.encode(assetSymbol))
+            : 0;
+        return bridgedAmount != 0
```
hayesgm, line 211: Weirdly also feels like `> 0` is a tad more natural for the intent here, though logically the same result

src/builder/QuarkBuilderBase.sol
```diff
@@ -407,25 +417,36 @@ contract QuarkBuilderBase {
...
         // Check if inserting wrapOrUnwrap action is necessary
         uint256 assetBalanceOnChain =
             Accounts.getBalanceOnChain(assetSymbol, chainId, chainAccountsList) + supplementalBalance;
         if (assetBalanceOnChain < amount && TokenWrapper.hasWrapperContract(chainId, assetSymbol)) {
-            // If the asset has a wrapper counterpart, wrap/unwrap the token to cover the transferIntent amount
+            // If the asset has a wrapper counterpart, wrap/unwrap the token to cover the amount needed for the intent
             string memory counterpartSymbol = TokenWrapper.getWrapperCounterpartSymbol(chainId, assetSymbol);
 
             // Wrap/unwrap the token to cover the amount
+            uint256 amountToWrap;
```
hayesgm, line 431: If we move to "all things wrapped maybe" then it might be nice that this logic simplifies a bit.

src/builder/QuarkBuilderBase.sol
```diff
@@ -407,25 +417,36 @@ contract QuarkBuilderBase {
...
+            if (useWrapUpTo) {
+                // If we are using the "wrap up to script", then the `amountToWrap` should be the entire amount needed
+                // for the intent
+                amountToWrap = amount;
+            } else {
+                // If we aren't using the "wrap up to" script, then we need to subtract the current balance from the
```
hayesgm, line 437: I'm trying to understand what's the serious case to not use `wrapUpTo`. Isn't it basically just a superior version of the same intent here?

src/builder/BridgeRoutes.sol
kevincheng96: It current reverts with `KeyNotFound`

src/builder/BridgeRoutes.sol
kevincheng96: Yeah I agree, I can change it

src/builder/QuarkBuilderBase.sol
kevincheng96: Agreed, having to adjust `amountToWrap` is kind of confusing. I'm probably going to bite the bullet and just change the script to `maybe wrap all`

src/builder/QuarkBuilderBase.sol
kevincheng96: In most cases, we know exactly how much needs to be wrapped so we don't need the extra overhead of `wrapUpTo`. But thinking about this again, that statement's not true. There can always be transactions happening outside of this Quark Operation that modifies the user's balance prior to execution of this operation. For this reason, I think we should just always default to `wrapUpTo`.

src/builder/BridgeRoutes.sol
kevincheng96: Done

src/builder/QuarkBuilderBase.sol
kevincheng96: Created the `wrap/unwrap all` variants for ETH and stETH. QuarkBuilder now always wraps/unwraps all.

src/builder/QuarkBuilderBase.sol
kevincheng96: Updated to always use `wrap/unwrap all`

src/WrapperScripts.sol
```diff
@@ -10,21 +10,21 @@ contract WrapperActions {
...
 
     function unwrapWETH(address weth, uint256 amount) external {
         IWETH(weth).withdraw(amount);
     }
 
-    function unwrapWETHUpTo(address weth, uint256 targetAmount) external {
-        uint256 currentBalance = address(this).balance;
-        if (currentBalance < targetAmount) {
-            IWETH(weth).withdraw(targetAmount - currentBalance);
+    function unwrapAllWETH(address weth) external payable {
```
hayesgm, line 24: Is unwrap really unwrapAll?

src/builder/BridgeRoutes.sol
hayesgm: Minor facepalm, but okay. I guess lack of generics and therefore default value

src/builder/QuarkBuilderBase.sol
```diff
@@ -37,7 +37,7 @@ contract QuarkBuilderBase {
 
     /* ===== Constants ===== */
 
-    string constant VERSION = "0.2.0";
+    string constant VERSION = "0.2.1";
```
hayesgm, line 40: Probably `0.3.0` at this point?

src/WrapperScripts.sol
kevincheng96: Yeah, should be right? I don't see how `unwrap` should be treated differently then `wrap` if we go the `wrap all` route.

kevincheng96: Based on the offline discussions we had, I changed the QuarkBuilder to now use `unwrapWETHUpTo` for WETH. Every other wrap/unwrap action is still using "all" instead of "up to".

PR #106, opened 2024-11-21
Add if to skip if either borrow/repay is having 0 amount

cwang25:
When just withdraw collateral, we should skip the borrow calls instead of calling borrow with 0 amount.
It looks like it was made in two scripts originally, but simple if in these two scripts do the tricks, so keep it in same script is more gas saving imo.


fluffywaffles (approved):
seems reasonable

